PR #16, opened 2014-12-11
new .executeAction() API for StoreMixin

zordius:
This patch wraps `this.props.context.executeAction()` to `this.executeAction()` to stop expose context location in code of components.

**Before:**
```javascript
// inside a component....
this.props.context.executeAction(xxxxxx);
```

**After:**
```javascript
// inside a component....
this.executeAction(xxxxxx);
```

When the context location moved to other place or the context solution changes, this will make all components safe from these change.

Possible context solutions:
https://github.com/brianmcd/contextify
https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html

### DIFF
--- a/mixins/StoreMixin.js
+++ b/mixins/StoreMixin.js
@@ -177,6 +177,18 @@ StoreMixin = {
             listener.store.removeChangeListener(listener.handler);
         });
         this.listeners = [];
+    },
+
+    /**
+     * wrap context executeAction
+     * @method executeAction
+     */
+    executeAction: function () {
+        if (!this.props.context) {
+            throw new Error('storeListener mixin was called but no context was provided for executeAction()');
+        }
+
+        return this.props.context.executeAction.apply(this.props.context, arguments);
     }
 };
 
--- a/tests/unit/mixins/StoreMixin.js
+++ b/tests/unit/mixins/StoreMixin.js
@@ -169,6 +169,20 @@ describe('StoreListenerMixin', function () {
     });
 
     describe('executeAction', function () {
+        it('should throw when no context provided', function () {
+            var Component = React.createClass({
+                mixins: [StoreMixin],
+                getInitialState: function () {
+                    this.executeAction(function () {},2,function(){});
+                    return {};
+                },
+                render: function () { return null; }
+            });
+           
+            expect(function () {
+                ReactTestUtils.renderIntoDocument(Component());
+            }).to.throw();
+        });
         it('should call context executeAction when context provided', function () {
             var Component = React.createClass({
                 mixins: [StoreMixin],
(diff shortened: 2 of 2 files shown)


gpbl:
I'm not sure if I like the idea. Explicitly referring `this.props.context` makes sure you are passing props correctly. If the reason is having a shortcut to `executeAction`, I don't think adding a mixin would make the thing more convenient. 

Though I also wonder @mridgway, what is the reason for not using react's contexts? Is because the naming issue (`this.context.context`), because they are not officially supported by React, or just a design choice?

zordius:
Maybe it's because the context support is in 'undocumented on purpose' state. refer to https://github.com/facebook/react/issues/580 , it may be removed in future.

So I propose 'hiding the context'...... if facebook remove native context support, then we can move to https://github.com/brianmcd/contextify , and mixin users do not need to do migration tasks when we do not expose context to them.

mridgway:
@gpbl Initially it was because we were unsure of Facebook's support for contexts. I think now that many of the most popular react libraries are using contexts, it should be safe to assume that `context` in some form will continue to exist in the future and we can start looking into using it.

zordius:
@mridgway it's a good news because then we can free components from 'always pass fluxible-app context to children' and make components more 'naturally compatible with other flux implements'.

gpbl:
@zordius OK sorry I didn't get the reasoning behind your PR :-) 

Anyway I see `this.props.context` (fluxible context) as a different thing from `this.context` (react context). Or is it safe to merge them? Fluxible contexts could get a different name, i.e. `this.context.fluxible`.

zordius:
`this.context.fluxible` is a good name.

My point is: if you change fluxible context name from `this.props.context` to `this.context.fluxible` , then all fluxible developers need to migrate their components from old code to new code. If we stop making user to access fluxible context directly, then them will not need to do this kind of migration works forever. So I PR this to propose a new API to hide the place of fluxible context.

gpbl:
I agree that this would help for the migration: but switching to react's contexts would anyway affect users not yet using the mixin.

I'm also for either add a new  `ActionMixin` mixin or rename your PR's `StoreMixin` to `FluxibleMixin` (I prefer the first).

gpbl:
Now that also the flux-router-component [is accepting](https://github.com/yahoo/flux-router-component/releases/tag/v0.5.5) react contexts, it would be awesome to have the StoreMixin using them as well (after digging into the code I've got the `context.fluxible` thing is not necessary, the context are just some functions).

mridgway:
I've adding support for executeAction and React contexts to #35. Closing this one as it should be completed when that PR is merged.

PR #25, opened 2015-01-13
[resolves #24] Rename package from fluxible-app to fluxible

### LINKED ISSUE #24: Rename `fluxible-app` to `fluxible`
See https://github.com/yahoo/flux-examples/issues/70. We'll maintain the current npm version when we rename and deprecate `fluxible-app` in npm.

mridgway:


yahoocla:
CLA is valid!

jedireza:
LGTM

redonkulus:
@mridgway do you want to rename [FluxibleApp.js](https://github.com/yahoo/fluxible/blob/master/lib/FluxibleApp.js) and the references inside it?

PR #52, opened 2015-02-26
Provide the current context as a child context.

joemcbride:
React.withContext is [deprecated in v0.13](https://github.com/facebook/react/blob/master/src/core/ReactContext.js#L54). This adds `getChildContext` to `FluxibleMixin` to help propagate the flux context to child components.

### DIFF
--- a/tests/unit/mixins/FluxibleMixin.js
+++ b/tests/unit/mixins/FluxibleMixin.js
@@ -41,6 +41,71 @@ describe('StoreListenerMixin', function () {
         delete global.navigator;
     });
 
+    describe('getChildContext', function(){
+        it('should return the context', function (done) {
+            var Component = React.createClass({
+                mixins: [FluxibleMixin],
+                statics: {
+                    storeListeners: [MockStore]
+                },
+                onChange: function () {
+                    done();
+                },
+                render: function () { return null; }
+            });
+            var Element = React.createFactory(Component);
+            var component = ReactTestUtils.renderIntoDocument(Element({context: context}));
+            var childContext = component.getChildContext();
+            expect(childContext).to.exist;
+            expect(childContext).to.be.an('object');
+            expect(childContext.getStore).to.be.an('function');
+            expect(childContext.executeAction).to.be.an('function');
+            mockStore.emitChange();
+        });
+        it('should propagate the context to child components', function (done) {
+            var childContext1, childContext2;
+            var ChildComponent1 = React.createClass({
+                mixins: [FluxibleMixin],
+                componentDidMount: function(){
+                    childContext1 = this.getChildContext();
+                },
+                render: function () { return null; }
+            });
+            var ChildComponent2 = React.createClass({
+                mixins: [FluxibleMixin],
+                componentDidMount: function(){
+                    childContext2 = this.getChildContext();
+                },
+                render: function () {
+                    var child = React.createFactory(ChildComponent1);
+                    return child();
+                }
+            });
+            var ParentComponent = React.createClass({
+                mixins: [FluxibleMixin],
+                statics: {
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

mridgway:
Nice. One question: should `getChildContext` also check to see if the context has already been passed in? For instance, if the `FluxibleMixin` is used in a child component, the child may not receive the prop, but it would receive the context from the top level component. Right now it looks like it would return undefined in `getChildContext`. Does that mean the children will get no context?

Wondering if it should be `return this.props.context || this.context`.

joemcbride:
>Wondering if it should be `return this.props.context || this.context`.

That sounds right, I'll add a test to confirm.

joemcbride:
Test confirmed the issue.  I'm not sure if there's a better way to grab to the child contexts?

joemcbride:
I also added in a version bump.

tests/unit/mixins/FluxibleMixin.js
```diff
@@ -54,6 +127,10 @@ describe('StoreListenerMixin', function () {
                 statics: {
                     storeListeners: [MockStore]
                 },
+                childContextTypes: {
```
mridgway: Shouldn't the FluxibleMixin provide this already? Why does it need to be declared in every test component?

joemcbride: React throws errors for public types on the context not declared in the `childContextTypes`.  This may be only for the top parent component.  Since the Mock context provides those additional public items they have to be declared or React pukes.

```
1) StoreListenerMixin getChildContext should propagate the context to child components:
     Error: Invariant Violation: ReactCompositeComponent.getChildContext(): key "dispatcher" is not defined in childContextTypes.
```

mridgway: Ah, I see. Makes sense.

mridgway:
:+1: Looks good to me. Great job!

We'll run some tests against our applications in the morning just to make sure this doesn't break any compatibility so we know if we need a minor version bump or not.

joemcbride:
:+1: Sounds good, that is a pretty big concern with how React treats those childContextTypes.  I haven't tacked on anything else to the context in my mini-app yet so I wouldn't see an issue.

mridgway:
Ok, so it looks like it will indeed conflict with other context types that are declared. I have found that if we just explicitly return the keys within this mixin in `getChildContext` then it should be safe: I opened a PR to your branch for you to look at and make sure it makes sense: https://github.com/joemcbride/fluxible/pull/1

joemcbride:
Taking a look.

joemcbride:
That'll work, and looks like you can declare you own mixin to propagate your own custom types if need be.

Do you want me to merge your pr or just force-push this one with those changes?

mridgway:
Yeah, that sounds good.

joemcbride:
I think we broke Travis CI.

mridgway:
Haha, it's an issue with contextify (jsdom dependency) in latest node@0.13 and io.js apparently. I think we're ok.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fluxible",
-  "version": "0.2.2",
+  "version": "0.2.3",
```
jedireza: Should we leave this out and bump during normal versioning?

mridgway: Ah, yes. @joemcbride can you remove this change? We prefer to do versioning separately in so that the tags and commits match up correctly.

joemcbride: @mridgway Removed.

mridgway:
:rocket:

jedireza:
:tada:

PR #53, opened 2015-02-27
Use optional displayName action attribute in debug logs

mbfisher:
Currently `FluxibleContext` uses `debug` to log when actions are executed with `executeAction`. It uses [`Function.name`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) to identify an action name to log. When code is minified, for example with UglifyJs's mangler, function names are removed and the log output contains an empty action name; `"Executing action  with payload"`. `Function.name` is read-only and so cannot forcibly be set outside of declaring a named function. 

This PR checks for an optional `displayName` property on action functions, which can be explicitly set, and is used in the log output if `name` is undefined.

```
// Declare named function - still useful.
function myAction(context, payload, done) {
    // Actiony stuffs
}

// Add displayName
myAction.displayName = 'myAction'
```

Existing behaviour should not be changed - if a `displayName` property is not the log output will still contain an empty name.

M

### DIFF
--- a/lib/FluxibleContext.js
+++ b/lib/FluxibleContext.js
@@ -82,8 +82,9 @@ FluxContext.prototype.plug = function (plugin) {
 FluxContext.prototype.executeAction = function executeAction(action, payload, done) {
     var self = this;
     payload = payload || {};
+    var displayName = action.name || action.displayName;
     setImmediate(function executeActionImmediate() {
-        debug('Executing action ' + (action.name) + ' with payload', payload);
+        debug('Executing action ' + displayName + ' with payload', payload);
         action(self.getActionContext(), payload, done);
     });
 };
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

mbfisher:
P.S. haven't added tests for this because, firstly, there doesn't look to be coverage of `executeAction`, and secondly because it changes logging behaviour which doesn't' affect the API.

M

mbfisher:
Found the `executeAction` test!

M

lib/FluxibleContext.js
```diff
@@ -82,8 +82,9 @@ FluxContext.prototype.plug = function (plugin) {
 FluxContext.prototype.executeAction = function executeAction(action, payload, done) {
     var self = this;
     payload = payload || {};
+    var displayName = action.name || action.displayName;
```
mridgway: I wonder if we should switch this precedence: `var displayName = action.displayName || action.name;`

mbfisher:
@mridgway Pushed!

M

mridgway:
Nice. Thanks!

mbfisher:
Follow up to lack of unit tests; this is in production for us now and working as expected.

M

PR #54, opened 2015-02-27
allow react 0.13 pre-releases

lingyan:
There is a request to relax react dependency to allow 0.13 pre-releases: https://github.com/yahoo/flux-router-component/issues/77#issuecomment-76346716

Wish I can use >=0.12.0 <=0.13.0, but can't, because this range does not satisfy react 0.13.0 beta and rc releases when using npm 2.0.

@mridgway @redonkulus  cc:/ @sAbakumoff

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "setimmediate": "^1.0.2"
   },
   "peerDependencies": {
-    "react": "^0.12.x"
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

redonkulus:
:shipit:

PR #59, opened 2015-03-09
disabled sudo to use docker images for faster builds

jedireza:
Ref: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/


yahoocla:
CLA is valid!

mridgway:
:boat:

PR #58, opened 2015-03-09
ability to provide an app level component action handler

jedireza:
Let's review and discuss the approach. When things solidify I'll add docs.

PR #60, opened 2015-03-10
bind the component action handler to the context instance

jedireza:
We missed this in #58 yesterday.


yahoocla:
CLA is valid!

mridgway:
It might be more clear to bind the `context` as a parameter instead of binding `this`.

jedireza:
Closing this PR. I'll open another with the changes we discussed IRL.

PR #63, opened 2015-03-12
Add a FluxibleComponent for wrapping app level component to provide child contexts

mridgway:
Also moves the `BaseStore` and `createStore` to an addons folder that could be required as a whole: `require('fluxible/addons').BaseStore`. I added some TODOs and comments for deprecating old `utils` path which we can do in the next minor version.

Also removed some unused test mocks.


yahoocla:
CLA is valid!

utils/BaseStore.js
```diff
@@ -1 +1,4 @@
-module.exports = require('dispatchr/utils/BaseStore');
+// @TODO deprecate
+//console.warn("require('fluxible/utils/createStore') is deprecated. Please use " +
```
jedireza: Uncomment warning?

utils/createStore.js
```diff
@@ -1 +1,4 @@
-module.exports = require('dispatchr/utils/createStore');
+// @TODO deprecate
+// console.warn("require('fluxible/utils/createStore') is deprecated. Please use " +
```
jedireza: Uncomment warning?

utils/BaseStore.js
mridgway: I'm thinking we should do this on next minor version.

jedireza: Oh I see. If we're going to do it, warnings won't hurt. And the sooner users see it, the better, right?

jedireza:
Can we share some details about what signifies an add-on? Why not have add-on for `Dispatcher` since it's already a hard dependency?

utils/BaseStore.js
mridgway: I learned the last time we added warnings that it can be annoying to suddenly see warnings without having done anything. I'd rather do it when the explicitly upgrade.

mridgway:
The idea is to keep it more inline with how React does it. Addons are things that you may use in production but don't get without explicitly requiring them. Utils are more build or test helpers.

utils/BaseStore.js
jedireza: Got it. Makes sense. What do you think about adding the version number, where the change is expected, to the todo comments? Like: 

```js
// @TODO deprecate in v0.6.x
// console.warn("require('fluxible/utils/createStore') is deprecated...
```

mridgway: I'm actually just going to do this through a Github milestone.

jedireza:
Thanks. Changes look good. :shipit:

redonkulus:
:+1:

utils/BaseStore.js
cesarandreu: :+1: on doing it with a Github milestone. Thanks for being considerate to library users.

klimashkin:
Is there some docs about FluxibleComponent? Is it about abandon ```withContext```?

mridgway:
I will add some docs to the readme. Since withContext is deprecated, the idea is that since mixins are not supported in ES6 components, the React team recommends using composition instead: wrapping your component and adding the child context so that the children can get access to the context.

klimashkin:
Hm, FluxibleComponent use childContextTypes, but in FluxibleMixin such comment exists about it https://github.com/yahoo/fluxible/blob/master/lib/FluxibleMixin.js#L47
I don't understand yet what is going on and what is planned -)

klimashkin:
Now I see. ```getChildContext``` in FluxibleMixin and in RouterMixin calls for every component and their children! Wow

What

klimashkin:
And another question. Why not in FluxibleComponent.js simplify method ```getChildContext``` to this?
```
getChildContext: function () {
    return this.props.context;
},
```

mridgway:
React throws an error if `getChildContext` returns more than what is defined in `childContextTypes`. Since plugins can add more to the component context, we have to defend against this error and only return the values that are defined in `childContextTypes`

mridgway:
I think the way forward is to remove the child context stuff from `FluxibleMixin` and encouraging the use of `FluxibleComponent` as a wrapper instead, but that would be a breaking change.

klimashkin:
It's absolutely right! Need to increment minor version and remove getChildContext from mixin!

PR #66, opened 2015-03-12
Lock down to 0.13; Document FluxibleComponent; Deprecate APIs

### LINKED ISSUE #65: react-dependency fails in older npm
hi,

if you use node-0.10, and the npm bundled with it, then this fails:

    npm install react@0.13.0 fluxible@0.2.9
    
    npm ERR! peerinvalid The package react does not satisfy its siblings' peerDependencies requirements!
    npm ERR! peerinvalid Peer fluxible@0.2.9 wants react@>=0.12.0 <=0.13.x

on node-0.12 it works.

im guessing this is a npm bug that was fixed in https://github.com/npm/node-semver/commit/c835c1ddef4a3d0e27c31fafa8098150815ef72b .

but, considering node-0.10 is still used, could fluxible perhaps use a different "backward" compatible dependency-version?

im not sure how exactly this works, but perhaps instead of "<=0.13.x" go with "<0.14" or something?

mridgway:
Resolves #65

 * Lock down to React 0.13.0 to give npm 1.x users a way to use React 0.13 (currently broken in Fluxible 0.2.x due to peerDependency range being broken in npm 1.x)
 * Deprecate APIs with warning messages:
   * `require('fluxible').Mixin` -> `require('fluxible').FluxibleMixin`
   * `require('fluxible/utils/BaseStore')` - > `require('fluxible/addons').BaseStore
   * `require('fluxible/utils/createStore')` - > `require('fluxible/addons').createStore
 * Document `FluxibleComponent`
 * `context.createElement()` will now wrap the app component with a `FluxibleComponent` to provide the child context correctly


yahoocla:
CLA is valid!

mridgway:
Depends on flux-router-component being bumped too. Working on that now.

redonkulus:
:shipit:

PR #71, opened 2015-03-19
Support promises in actions

### LINKED ISSUE #69: Support Promises in actions
Support handling promises that are returned from actions instead of using callbacks.

cesarandreu:
Closes #69.

I added lots of tests to executeAction. New behavior:

* Execute action returns a promise
* If done callback is passed to execute action, it'll be called when the promise is resolved or rejected
* If action returns a promise, resolve/reject execute action promise with result/error
* If action takes less than 3 params, execute action promise will be resolved with action return value
* If action throws an error, execute action promise will be rejected with thrown value
* When action takes 3 params, it'll receive a callback that can reject/resolve the execute action promise

Should be almost entirely backwards compatible. While writing this I realized that my change to `componentAction.executeAction` actually breaks backwards compatibility with component action handler, since now it'll only get executed when there's errors. Should I revert this change? 

* I'm not sure what your styleguide is like... I tried to replicate existing code and follow the linter's suggestions.
* Should I update the docs as well?
* Is my jsdoc-style comment alright? I'm not sure how to explain the different branches, so I just added a few lines on top of the method.
* Is the goal to keep both promises and callbacks or are you gonna deprecate callbacks? (I'd vote for deprecating em)


yahoocla:
CLA is valid!

mridgway:
This looks really good!

> Should be almost entirely backwards compatible. While writing this I realized that my change to componentAction.executeAction actually breaks backwards compatibility with component action handler, since now it'll only get executed when there's errors. Should I revert this change?

I'm ok with it just being called for error cases. That was the primary reason that handler was added.

> I'm not sure what your styleguide is like... I tried to replicate existing code and follow the linter's suggestions.

The code and comments look just fine ;)

> Should I update the docs as well?

I will be pushing some significant docs changes in the next 24 hours, so it may be good to hold off on that until then. I will let you know when these changes are in so you can update the PR.

> Is my jsdoc-style comment alright? I'm not sure how to explain the different branches, so I just added a few lines on top of the method.

Looks fine to me as well.

> Is the goal to keep both promises and callbacks or are you gonna deprecate callbacks? (I'd vote for deprecating em)

This is something we'll have to figure out in the next few releases. I've always been wary of using Promises due to how they use try/catch, but since the internals are using Promises regardless, it may make sense to just use them without the callbacks. So this is still up in the air.


I want to play some more with the code tomorrow to make sure we don't get any silent failure cases like the Promises eating throws. So with that and the documentation updates, we'll be good to go.

cesarandreu:
^ @mridgway, I wrote up some docs; any feedback or suggestions?

jwalton:
Instead of using bluebird, I'd use [es6-promise](https://github.com/jakearchibald/es6-promise).  It's 1/13th the size, minified, and supports the full ES6 Promise specification.

What I did in [lol-js](https://github.com/jwalton/lol-js/blob/master/src/client.coffee#L32-L33) was default to es6-promise, but I accept a Promise constructor as an option so the end-user can set whatever Promise/A+ compliant library they want to use.

jwalton:
Also, BTW, +1 for this.  Let's make this happen.  :D

mridgway:
@jwalton Yeah, I think we will switch it to es6-promise, I didn't realize how big bluebird got when bundling it. One pattern we've used before is `var PromiseLib = global.Promise || require('es6-promise');` which would allow overriding globally across the application.

mridgway:
@cesarandreu This looks good :+1: . I'm looking to pull this in today.

cesarandreu:
Sure thing :D. Sounds good. :shipit:

mridgway:
Many thanks @cesarandreu! I have some other tasks to finish up for the 0.4.0 release, but will try to get this done ASAP.

cesarandreu:
Great, thanks!

bruderstein:
Just a note - I'm pretty sure this has had the side effect of making executeAction synchronous. So if you call executeAction following a store change event, you end up with a dispatchr error now (dispatch attempted whilst dispatch in progress).  This is obviously not a recommended behaviour, so (IMHO) it's an improvement, but previously it would be silent due to the setImmediate in executeAction().

It might be good to add this note to the CHANGELOG, for anyone else seeing this issue.  I had some hidden cascading updates, and wondered what had changed to cause them since upgrading.

mridgway:
@bruderstein Yes this was a regression. We are tracking this in https://github.com/yahoo/fluxible/issues/139

PR #78, opened 2015-03-24
[resolves #73] Remove requirement of using createFactory before passing ...

### LINKED ISSUE #73: appComponent and React.createFactory
I've recently updated from 0.1.5 and received a strange error message from React, specifically `ReactClass#createClass`:

```
Uncaught TypeError: Cannot read property '__reactAutoBindMap' of null
```

I fixed this by wrapping my `appComponent` in `React.createFactory`:

```js
var App = require('App.jsx');
var app = new Fluxible({
    component: React.createFactory(App)
});
```

I have no idea why it didn't work nor why it now does! Do you understand what's happening? Have you had any other reports of this?

M

mridgway:
...component to Fluxible instance

Next minor can probably deprecate passing a factory in since it will not be need when we use `React.createElement()`

### DIFF
--- a/lib/FluxibleContext.js
+++ b/lib/FluxibleContext.js
@@ -46,9 +46,16 @@ FluxContext.prototype.createElement = function createElement(props) {
     if (!Component) {
         throw new Error('A component was not specified.');
     }
+    var componentInstance;
+    if (!Component.hasOwnProperty('prototype')) {
+        // TODO: add warning to deprecate factories
+        componentInstance = Component(props);
+    } else {
+        componentInstance = React.createElement(Component, props);
+    }
     return FluxibleComponent({
         context: this.getComponentContext()
-    }, Component(props));
+    }, componentInstance);
 };
 
 /**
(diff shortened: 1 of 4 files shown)


yahoocla:
CLA is valid!

lib/FluxibleContext.js
```diff
@@ -46,9 +46,16 @@ FluxContext.prototype.createElement = function createElement(props) {
     if (!Component) {
         throw new Error('A component was not specified.');
     }
+    var componentInstance;
+    if (!Component.hasOwnProperty('prototype')) {
+        // TODO: add warning to deprecate factories
```
Vijar: Why not just console.warn now?

tests/unit/lib/FluxibleContext.js
```diff
@@ -20,7 +20,7 @@ describe('FluxibleContext', function () {
 
     describe('createElement', function () {
         it('should receive the correct props and context', function (done) {
-            var Component = React.createFactory(React.createClass({
+            var Component = React.createClass({
```
lingyan: I would keep the old test, until the createFactory support is removed.

mridgway: Re-added.

lib/FluxibleContext.js
mridgway: Not until next minor release.

redonkulus:
:shipit:

lingyan:
:+1:

PR #92, opened 2015-03-30
convert examples to es6

redonkulus:
@mridgway @jedireza @Vijar 

First pass at our migration to ES6. Figured I'd do the easy part first and update the docs.


yahoocla:
CLA is valid!

cesarandreu:
You guys thinking of moving to using babel for the codebase as well?

Vijar:
@cesarandreu yep! We're just starting with examples first.

Converting the codebase would be a cool project, but it would be compiled back into es5 anyway. But with the examples and docs, we the get benefit of users seeing how they can use fluxible with their es6 projects.

PR #93, opened 2015-03-30
Allow context plugins rehydrate to be async

mridgway:
Same implementation as what app level plugins use. Required for contextual loading of assets, ala webpack lazy requires.


yahoocla:
CLA is valid!

redonkulus:
:shipit:

PR #94, opened 2015-03-30
[0.2.x] Allow context plugins rehydrate to be async

mridgway:
Backported to 0.2.x branch.


yahoocla:
CLA is valid!

redonkulus:
:shipit:

PR #97, opened 2015-03-31
[resolves #89] executeAction no longer coerces falsy values to an object

### LINKED ISSUE #89: Unintentional Value Coercion
In `FluxContext.prototype.executeAction` there is the following line: `payload = payload || {};` [Source](https://github.com/yahoo/fluxible/blob/master/lib/FluxibleContext.js#L93). If you pass `false` as the `payload`, then your payload silently gets changed into an object. Is the intent that `payload` is ALWAYS an object? If so, I think there should be a thrown error if it is not an object. If it doesn't have to be an object, then the check needs to be updated.

Looking at the code, it appears that `payload` is supposed to be an object but I'm curious why that is? If I want to store something like visibility, I would only need to be passing `true` or `false`. What are the thoughts on the intent of `payload` in the fluxible context?

mridgway:


yahoocla:
CLA is valid!

jedireza:
:+1:

PR #98, opened 2015-04-01
[extension of #71] Support promises in actions

mridgway:
* switched to es6-promise for file size reasons
* added test and fix for callback errors being swallowed by internal Promise

Found an interesting quirk with v8's Promises and domains which is why the tests use es6-promise instead of native.


yahoocla:
CLA is valid!

tests/unit/lib/FluxibleContext.js
```diff
@@ -86,11 +108,132 @@ describe('FluxibleContext', function () {
...
+                    expect(executeActionError).to.equal(err);
+                    done();
+                });
+            });
+
+            it('should not swallow callback errors', function (done) {
+                // Error is expected, but will not be catchable. Crudely using domain.
+                var d = domain.create();
+                d.on('error', function () {
+                    console.log('test');
```
lingyan: remove console.log, and maybe assert the error is the `test` error that was thrown.

lingyan:
Left a minor comment.  Otherwise looks good.

cesarandreu:
\o/

PR #99, opened 2015-04-01
Update to dispatchr@0.3.x and change test utils to use new-less API

mridgway:
You can see the new test utils APIs in the doc updates.

```js
var MockComponentContext = require('fluxible/utils/MockComponentContext')();
MockComponentContext.registerStore(FooStore);
var context = new MockComponentContext();
```

becomes

```js
var createMockComponentContext = require('fluxible/utils').createMockComponentContext;
var context = createMockComponentContext({
    stores: [FooStore]
});
```

### DIFF
--- a/addons/BaseStore.js
+++ b/addons/BaseStore.js
@@ -1,1 +1,1 @@
-module.exports = require('dispatchr/utils/BaseStore');
+module.exports = require('dispatchr/addons/BaseStore');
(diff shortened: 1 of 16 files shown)


yahoocla:
CLA is valid!

jedireza:
:shipit: on a :ship:

Code looks great BTW

3den:
:+1:

PR #101, opened 2015-04-01
[resolves #100] Add component factory deprecation; Remove APIs that were...

### LINKED ISSUE #100: Deprecate passing factories into Fluxible constructor
Previously it was required to pass in a factory to Fluxible's `component` option, but it is no longer necessary since we are using `React.createElement`.

mridgway:
... deprecated in previous version


yahoocla:
CLA is valid!

jedireza:
:+1:

PR #102, opened 2015-04-01
Update libraries.md

jwalton:


yahoocla:
CLA is valid!

jedireza:
Nice. What are your thoughts about renaming the module to `fluxible-plugin-service-proxy`. The hope is to have a more standard naming convention for plugins. We'll be adding a note about this to the docs soon.

jwalton:
I named it something similar originally, but that's pretty wordy.  :)  I can do that, though.

jwalton:
Is there a sane way to rename a package on npm?  Or do I just unpublish the old one?

jedireza:
For longer lived modules I think [`$ npm deprecate ...`](https://docs.npmjs.com/cli/deprecate) is a good idea.

 > This command will update the npm registry entry for a package, providing a deprecation warning to all who attempt to install it.

Otherwise yes, unpublishing current and publish new should do the trick.

jedireza:
I'd like to see what others think about this.

cc// @redonkulus @Vijar @mridgway @lingyan

jwalton:
Well, I've *just* unpublished it.  :P  So I'll go ahead and rename it.

jedireza:
Ok cool. I think it's a good move.

jwalton:
I also did [this bit of cleverness](https://github.com/benbria/node-fluxible-plugin-service-proxy/blob/master/index.js#L29-L36) so if someone else defines a `getService()`, mine and theirs will live together in peaceful coexistence.  :)

Vijar:
I think we should stick with the convention "fluxible-plugin-*" to make it clear this is a plugin for fluxible as opposed to something like ["fluxible-action-utils"](https://github.com/yahoo/fluxible-action-utils) which isn't a plugin but is still related to fluxible.

jedireza:
PR LGTM :shipit:

Vijar:
:ship:

PR #106, opened 2015-04-01
[docs] fix improper merge conflict resolution from #99

koulmomo:
@redonkulus @mridgway


yahoocla:
CLA is valid!

jedireza:
:+1: good catch

mridgway:
Yeah, I'm fixing this now. There's other issues in there too... the code doesn't work.

mridgway:
Fixed in 41218981968ab6bad31562155c8d23fb8f023847

PR #108, opened 2015-04-02
Example Link for Introducing the Fluxible Library

jaxoncreed:
I found this repo to be very helpful when converting my application from simply isomorphic-flux to one that used Fluxible. Having a link to it would be convenient.


yahoocla:

yahoocla:
CLA is valid!

redonkulus:
Thanks!

PR #109, opened 2015-04-02
FluxibleContext.executeAction() returns the promise now

koalabruder:
If you have an action that should return a promise it is essentiell that
FluxibleContext.executeAction() returns your promise. Thats the fix.


yahoocla:

yahoocla:
CLA is valid!

mridgway:
Hi @koalabruder, the promise is not returned because Component `executeAction` should be fire-and-forget.

Issue #79, opened 2015-03-24
Namespacing plugins


jedireza:
Currently plugins add functionality to the root of `context`. This makes it easy for plugins to pollute `context` or collide features with one another.

Also in larger apps it can get hard to know which plugin is responsible for a specific piece of functionality. I'd like to see plugin functionality to be namespaced. `context.pluginName.someFeature(...)` as an example.

Plugin names should be unique and we should throw during `app.plug({ ... })` if a a plugin with the same name has already been registered.

If you take a look at `react-router@0.13.x`, they're kind of following a similar idea in that `context.router` contains all their functionality.

mridgway:
Sounds reasonable to me. Would love to hear more about how people are using plugins in the wild.

mridgway:
plug*Context() functions would now return an object that contains their functions rather than modifying the context object directly.

cesarandreu:
I'm currently using a plugin to allow promises in actions, another to add a wrapped fetch (so I can add stuff like csrf token to all requests getting sent, and overwrite how the csrf token is taken depending on environment), and another to abstract react-router and add the instance to my action context. 

Something I like about the current approach is I can monkey-patch stuff (e.g. executeAction).

I haven't built anything big enough with Fluxible such that this would be a problem, though.

alexaivars:
Currently I don't se the need of having that many plugins that this would result in a problem. Currently I'm working on a few big app but only using plugins for easy access to analytics tracking, a remote service interface in actions and sharing environment variables between the server and client.

jedireza:
Should we close this one too?

redonkulus:
@jedireza  :+1:

PR #110, opened 2015-04-03
[wip] initial attempt to namespace plugins

jedireza:
This is my initial attempt to namespace plugins and address #79.

One goal I had was to ensure that plugin authors don't need to change anything. However when we used to pass the context to a plugin's `plugActionContext`, `plugComponentContext` or `plugStoreContext`, it may have already been decorated with some resources.

For example in the `getActionContext` method we were creating this context object:

```js
var actionContext = {
    dispatch: this._dispatcher.dispatch.bind(this._dispatcher),
    executeAction: this.executeAction.bind(this),
    getStore: this._dispatcher.getStore.bind(this._dispatcher)
};
```

and passing that to the plugin's `plugActionContext` function, which it then decoratea with it's own functionality. But now we're passing an empty object `{}` just for that plugin to decorate. My concern is... what if plugin authors were using the resources on the context we were passing.

If the plugins do indeed need those resources:

 - We could include copies of the decorated resources (`dispatch`, `executeAction` and `getStore` in the example above) to each plugin. No changes are required for plugin authors _but_ it feels sloppy to have multiple references to these resources floating around.
 - We could to pass the ~~parent~~ real context as another argument to their `plugActionContext`, `plugComponentContext` and `plugStoreContext` functions. This _would_ require plugin authors to change code.
 - If we accept that changes may be required by plugin authors... we do already pass a reference to `this` as the second argument, plugin authors could access the resources they need from there.
 - And another option I almost forgot (mentioned in #79) is that we could require a new object be returned from these `plug*Context` functions. This approach may be the easier change to implement for plugin authors.

Your thoughts are welcome.

### DIFF
--- a/lib/Fluxible.js
+++ b/lib/Fluxible.js
@@ -8,7 +8,6 @@ var debug = require('debug')('Fluxible');
 var async = require('async');
 var FluxibleContext = require('./FluxibleContext');
 var dispatchr = require('dispatchr');
-var React = require('react');
 
 /**
  * Provides a structured way of registering an application's configuration and
@@ -34,7 +33,14 @@ function Fluxible(options) {
     // Options
     this._component = options.component;
     this._componentActionHandler = options.componentActionHandler || this._defaultComponentActionHandler;
-    this._plugins = [];
+    this._plugins = {};
+
+    if (options.plugins) {
+        var self = this;
+        options.plugins.forEach(function initRegisterPlugins(plugin) {
+            self.plug(plugin);
+        });
+    }
 
     if (!this._component.hasOwnProperty('prototype')) {
         console.warn('It is no longer necessary to wrap your component with ' +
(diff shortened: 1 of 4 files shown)


yahoocla:
CLA is valid!

cesarandreu:
I think it'd be valuable to have some way of plugging the real context, and not the namespaced context. 

If you can't plug the real context, then you won't be able to provide functions directly like with getStore, dispatch, or executeAction. You also can't overwrite / wrap their functionality. For example, I did this with executeAction so it would handle promises, before that change got merged.

jedireza:
@cesarandreu thanks for sharing. I hadn't considered that use case.

mridgway:
I don't think we're going to namespace plugins at this point since it would limit the control over the contexts that allowed things like fluxible-plugin-promises.

cesarandreu:
It might be worth writing this as a suggestion in the docs.

For example, I have a `RequestPlugin` which adds `actionContext.executeRequest` to handle most of my request. However, I also expose the internals (i.e. `fetch`),  under `actionContext.request = {...}`, in case I need to use the internal abstractions for some reason. (Because apps are a huge pain and you always wind up finding edge-cases once your app grows to a certain size :(, so having an escape hatch is convenient.)

PR #113, opened 2015-04-04
Should not error when no component is passed

cesarandreu:
The component option in the Fluxible constructor is optional, so it shouldn't throw errors.


yahoocla:
CLA is valid!

mridgway:
Good catch! I think I may move this warning into `context.createElement` since users may still be using `component` without `createElement`.

cesarandreu:
That would be even better! I'm one of those people :D.

PR #114, opened 2015-04-04
Fix inline documentation

cesarandreu:
Inline docs were outdated.


yahoocla:
CLA is valid!

mridgway:
Thanks!

PR #115, opened 2015-04-04
Move factory warning to context.createElement

mridgway:
This warning is really only relevant when user is using `context.createElement` since they may still want a factory if they call `app.getComponent()`.

@cesarandreu @jhwang


yahoocla:
CLA is valid!

cesarandreu:
:shipit:!

Could you do a release once it's merged?

mridgway:
Yep. Will do now.

PR #116, opened 2015-04-04
Expose the reduce happening in connectToStores so that store state can be merged

mridgway:
**Breaking Change** - I don't think there is a way to make this work with the existing implementation from what I can find.

This allows the use case where state from two stores needs to be merged together, for instance `FeaturedStore` with a list of IDs and `PhotoStore` where the individual items can be retrieved:

```js
connectToStores(Component, [FeaturedStore, PhotoStore], {
    FeaturedStore: function (state, store, props) {
        state.featuredIDs = store.getFeaturedIDs();
    },
    PhotoStore: function (state, store, props) {
        state.photos = value.featuredIDs.map(function (id) {
            return store.getPhotoById(id);
        });
    }
})
```

@gpbl

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ $ npm install --save fluxible
 import Fluxible from 'fluxible';
 import React from 'react';
 import {connectToStores, createStore, provideContext} from 'fluxible/addons';
+import objectAssign from 'object-assign';
 
 // Action
 const action = (actionContext, payload) => {
@@ -75,8 +76,8 @@ class App extends React.Component {
 }
 
 App = provideContext(connectToStores(App, [FooStore], {
-    FooStore(store) {
-        return store.getState();
+    FooStore(state, store) {
+        objectAssign(state, store.getState());
     }
 }));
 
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

mridgway:
Other ways to handle this would be allowing the last params in `connectToStores` be a function that receives all of the stores:

```js
connectToStores(Component, [FeaturedStore, PhotoStore], function (stores, props) {
  var featuredIDs = stores[FeaturedStore].getFeaturedIDs();
  
  var photos = featuredIDs.map(function (id) {
    return stores[PhotoStore].getPhotoById(id);
  });
  return photos;
});
```

or

```js
connectToStores(Component, [FeaturedStore, PhotoStore], function (featuredStore, photoStore, props) {
  var featuredIDs = featuredStore.getFeaturedIDs();
  
  var photos = featuredIDs.map(function (id) {
    return photoStore.getPhotoById(id);
  });
  return photos;
});
```

I lean more towards the first version with the store hash instead variable arguments.

mridgway:
Moving to an issue for further discussion.

PR #119, opened 2015-04-05
Check error object in component action handler test

cesarandreu:
Small improvement to component action handler test.
It makes sure the thrown action error is the error passed to component action handler.


yahoocla:
CLA is valid!

PR #120, opened 2015-04-05
Component action handler errors should be thrown

cesarandreu:
In my promises PR I hadn't realized there was still the potential for swallowed errors. I stumbled upon this problem today while playing around with the new release. :(

If you execute a component action and it throws an error, it'll call component action handler. However, if component action handler throws the error again, you won't be notified.

With this update, if component action handler throws an error, it'll be thrown as expected. I think this behavior is more sensible.

Testing this is a bit hacky, but I don't know if there's a better way.


yahoocla:
CLA is valid!

cesarandreu:
GitHub isn't updating the commits, so I'm gonna close it and open it again.

PR #124, opened 2015-04-07
connectToStores: componentWillReceiveProps support and context access for change handlers

Ambroos:
React-Router triggers a componentWillReceiveProps when you stay within the same route but the params or queries change (for example, navigating from one todo directly to another). Without componentWillReceiveProps in the route handlers (the wrapped components in this case) you can't get new data from the store when this happens.

The second change in the second commit isn't necessary but felt like the logical next step. It allows you to pass in contextTypes to access from the stateGetter(s). React-Router v0.13 uses context as well, and the only way to access route params/queries is through the component context. Without access to context you either have to always just get the whole store state and do the 'selecting' in your component, but that doesn't feel like the right place to do those things in this architecture.

What do you think?

Example of context usage:
```javascript
Component = connectToStores(Component, [TodoStore], (stores, props, context) => {
    let params = context.router.getCurrentParams();
    return {
        todo: stores.ToDoStore.get(params.id)
    };
}, {
    router: React.PropTypes.func.isRequired
});
```

(Also corrected two small things in the process, props were passed through to `getStateFromStores` in `getInitialState` but there was no parameter for it, and `return StoreConnector;` didn't have a semicolon.)

### DIFF
--- a/addons/connectToStores.js
+++ b/addons/connectToStores.js
@@ -19,13 +19,14 @@ var contextTypes = require('../lib/contextTypes');
  *      the full state object. Receives `stores` hash and component `props` as arguments
  * @returns {React.Component}
  */
-module.exports = function connectToStores(Component, stores, getStateFromStores) {
+module.exports = function connectToStores(Component, stores, getStateFromStores, extraContextTypes) {
     var componentName = Component.displayName || Component.name;
+    var componentContextTypes = objectAssign({
+        getStore: contextTypes.getStore
+    }, extraContextTypes);
     var StoreConnector = React.createClass({
         displayName: componentName + 'StoreConnector',
-        contextTypes: {
-            getStore: contextTypes.getStore
-        },
+        contextTypes: componentContextTypes,
         getInitialState: function getInitialState() {
             return this.getStateFromStores();
         },
@@ -49,7 +50,7 @@ module.exports = function connectToStores(Component, stores, getStateFromStores)
                     var storeName = store.name || store.storeName || store;
                     storeInstances[storeName] = this.context.getStore(store);
                 }, this);
-                return getStateFromStores(storeInstances, overrideProps || this.props);
+                return getStateFromStores(storeInstances, overrideProps || this.props, this.context);
             }
             var state = {};
             //@TODO deprecate?
--- a/addons/connectToStores.js
+++ b/addons/connectToStores.js
@@ -27,33 +27,36 @@ module.exports = function connectToStores(Component, stores, getStateFromStores)
             getStore: contextTypes.getStore
         },
         getInitialState: function getInitialState() {
-            return this.getStateFromStores(this.props);
+            return this.getStateFromStores();
         },
         componentDidMount: function componentDidMount() {
             stores.forEach(function storesEach(Store) {
                 this.context.getStore(Store).addChangeListener(this._onStoreChange);
             }, this);
         },
+        componentWillReceiveProps: function(nextProps) {
+            this.setState(this.getStateFromStores(nextProps));
+        },
         componentWillUnmount: function componentWillUnmount() {
             stores.forEach(function storesEach(Store) {
                 this.context.getStore(Store).removeChangeListener(this._onStoreChange);
             }, this);
         },
-        getStateFromStores: function () {
+        getStateFromStores: function (overrideProps) {
             if ('function' === typeof getStateFromStores) {
                 var storeInstances = {};
                 stores.forEach(function (store) {
                     var storeName = store.name || store.storeName || store;
                     storeInstances[storeName] = this.context.getStore(store);
                 }, this);
-                return getStateFromStores(storeInstances, this.props);
+                return getStateFromStores(storeInstances, overrideProps || this.props);
             }
             var state = {};
             //@TODO deprecate?
             Object.keys(getStateFromStores).forEach(function (storeName) {
                 var stateGetter = getStateFromStores[storeName];
                 var store = this.context.getStore(storeName);
-                objectAssign(state, stateGetter(store, this.props));
+                objectAssign(state, stateGetter(store, overrideProps || this.props));
             }, this);
             return state;
         },
@@ -65,5 +68,5 @@ module.exports = function connectToStores(Component, stores, getStateFromStores)
         }
     });
 
-    return StoreConnector
+    return StoreConnector;
 };


yahoocla:
CLA is valid!

mridgway:
:+1: this looks good to me. I want to think about whether we're going to have any new parameters for `connectToStores` in the future so we don't keep adding them on. If we have one more, we should probably start an `options` parameter. Not sure we're there yet though.

gpbl:
Having `context` directly passed to the getter makes its `store` argument less useful. Then we could simplify the API even more:

```js
connectToStores(Component, ['MyStore'], (context, props) => 
   ({ context.getStore('MyStore').getState()} )
)
```

which is not so different from the actual implementation.

```js
connectToStores(Component, ['MyStore'], (stores, props) => 
    ({ stores['MyStore'].getState() })
);
```

yet it gives more freedom. For example, when we won't listen to *all* the stores, it would become:

```js
connectToStores(Component, ['MyStore'], (context, props) => {
  const foo = context.getStore('MyStore').get('foo');
  return {
    otherStuff: context.getStore('MyOtherStore').get({ bar: foo} )
})
```

Going further with the idea, if I won't listen to anything, i pass just two arguments but it is not a `connectToStores` anymore :-)

```js
connectToStores(Component, (context, props) => context.getStore('MyOtherStore').getState())
```

I wonder if it is worth to take this sort of implementation out of fluxible...

/cc @koulmomo

Ambroos:
That is not a bad idea, although you have to be careful not to start any actions that do immediate dispatches inside connectToStores. We still need `contextTypes` too.

How about:

`connectToStores(Component, [ Store ], (context, props) => {}, contextTypes = { getStore: ... })`

contextTypes is optional, and by default we only provide `getStore`. This allows most people to just use it without the last parameter.

When people try to request executeAction from contextTypes, we should throw a warning? I've noticed that since Fluxible v0.4 it's no longer possible to dispatch inside another dispatch, which is something that happens a lot if you do `executeAction` inside a dispatch, and the onChange callback of `connectToStores`.

I'll throw an implementation of this on NPM. Should I update this pull request too for a next major version, since this would be a breaking change?

mridgway:
I'd like to make this change in 0.5. We'll just pass the `context` in instead of the `stores` object.

pumpikano:
:+1: , I'd love to start using this in an alpha of 0.5

gpbl:
@mridgway I'm not sure I understood the plans for closing this issue :-) 

Use case: I have a component connecting to a "list store", i.e. a store with a list of identities referring to another store. To cross data between stores, with the actual `connectToStore` implementation I'm forced to make my component "context aware", for example [this one](https://github.com/gpbl/isomorphic500/blob/dd44a69d26b8f6b9d640b03d0fe6cbd3008e13f4/src/pages/FeaturedPage.js#L22).

If the plan is to pass directly `context` as callback's argument of `connectToStore` (as in [my comment](https://github.com/yahoo/fluxible/pull/124#issuecomment-90829314) above) I could fix this and keep my component "dumb", and remove references to a `context`.

Is this the plan?

(not sure I understand why we should pass `contextTypes` as well, @Ambroos)

Ambroos:
@gpbl Since the wrapper component is still just a React component, it only has access to the parts of the context it requests with `contextTypes`. I'm not sure how easy it is to somehow get the full context object, but it may not be a bad idea to just enforce the contextTypes requirement (like everywhere else in React).

gpbl:
thanks @Ambroos now I see: basically, a way to pass down context elements other than fluxible's ones. I don't get either your suggestion for a warning to `executeAction`, why shouldn't it be used – for example, after a user's interaction?

A [previous version](https://github.com/gpbl/isomorphic500/blob/5f43636d8dc72d3991e67c8ed58721f16e2cc0a8/src/pages/FeaturedPage.js#L33) of that component was listening to both stores. The first store is a store containing only ids, the other contains the actual objects. A single action will update both stores, however the page should render when the first store emit a change, and only when the second store contains the referred objects (so I'm using `waitFor` [in the first store](https://github.com/gpbl/isomorphic500/blob/master/src/stores/FeaturedStore.js#L25)). I don't need to listen to the second store, but just the first – otherwise the component would render twice – with the risk to have the two stores not synched. (phew I hope i'm being clear 😅).

If a store depends from another, with the actual implementation we cannot use `connectToStore` to listen the first store and only have access to the second one. For doing that I need to add `contextTypes` to the component.

Ambroos:
@gpbl The warning on using `executeAction` should happen only inside `connectToStores`. You should not be executing actions from inside it, to avoid cascading updates. Basically, `context` inside `connectToStores` should never have access to `executeAction`, that's it.

I get what you mean with the rendering twice issue. I tend to solve this with `shouldComponentUpdate`. If  your function `PhotoStore.getMultiple` returns an immutable object that's always the same (or you just write a fast equals check), you're good. Your component will receive the `new` props, but decide that it doesn't need to rerender. With your current solution you 'rely' on `FeaturedStore` to wait for the others, which may not work in the long run. I'm more of a fan of how you handled it in the previous version.

But, even if you wanted to keep your logic of only listening to `FeaturedStore` (nothing wrong with that), you could still use the 'new' way of doing things without needing context inside your component:

```javascript
class FeaturedPage extends React.Component {

  static propTypes = {
    photos: PropTypes.array.isRequired
  }

  render() {
    const { photos } = this.props;
    return (
      <div>

        <div className="ThumbnailCollection">
          {
            photos.map(photo => <Thumbnail key={photo.id} size="small" photo={photo} />)
          }
        </div>

      </div>
    );
  }

}

FeaturedPage = connectToStores(FeaturedPage, [FeaturedStore], (context) => {
  return {
    photos: context.getStore(PhotoStore).getMultiple(ids)
  };
}, {getStore: React.propTypes.func});
```

I think that way your logic is split better. Getting data from the stores is something that happens in `connectToStores`. You component itself can focus on rendering and handling user interaction instead, plus you don't have to give your component access to the context. What do you think?

gpbl:
Alright yes, this is also the API I'm waiting for ! 👍

Then I wonder why passing `context` to the callback? Shouldn't be safer to pass just *all* the registered stores?
Example:
```js
FeaturedPage = connectToStores(FeaturedPage, ["FeaturedStore"], (stores, props) => {
  const ids = stores["FeaturedStore"].getIdsForFeature(props.feature)
  return {
    photos: stores["PhotoStore"].getMultiple(ids) // note PhotoStore wasn't being listened
  };
});
```
Benefits:
* no need for a third argument to restrict `context`
* full access to any store (also those not being listened)
* it's compatible with the current implementation

Ambroos:
@gpbl It's a start, but not really versatile enough I think. I use React-Router in my current project (in too deep to switch), and you need context access to get data from the routing state. By allowing full context access in `connectToStores`, I can check route params inside it and get the right data from the stores. And, besides that, there are many useful things you can do with context, it's a pretty cool thing.

gpbl:
you are right, then let's hope for this change in `0.5`!

jeron-diovis:
Completely agreed with passing context to state getter. 

But why need to [remove](https://github.com/yahoo/fluxible-addons-react/blob/master/connectToStores.js#L53-L54) [old interface](https://github.com/yahoo/fluxible-addons-react/blob/master/connectToStores.js#L78)? 

It's just about convenience. When I need several props from single store in my component, with new API I should write all these things:
```js
connectToStores(..., (context) => {
    const store = context.getStore('MyStore');
    return {
        prop: store.prop,
        another_prop: store.another_prop
    }
});
```
while with old one it can be just
```js
connectToStores(..., ({ MyStore: { prop, another_prop } }) => ({ prop, another_prop });
```

Of course, potentially, with `mergedContext` some store instances can overridden by props from context, but is it a so realistic usecase?
So why not to keep merged context?

PR #126, opened 2015-04-08
Fix typo in quick-start.md

juandopazo:
Just so.


yahoocla:
CLA is valid!

mridgway:
Thanks!

PR #131, opened 2015-04-16
Update outdated connectToStores docs

cesarandreu:
Not sure if should just update docs with getStateFromStores function and skip mentioning storeGetters, or if storeGetters should be mentioned (since it's not (deprecated yet)[https://github.com/yahoo/fluxible/blob/master/addons/connectToStores.js#L52]).

But it's weird that the example below this is using the getStateFromStores function instead of storeGetters.


yahoocla:
CLA is valid!

mridgway:
Good catch. Yes, `storeGetters` is undocumented since it will likely be deprecated. Thanks!

PR #132, opened 2015-04-17
Make context.executeAction return executeActionPromise

geta6:
According to [document](http://fluxible.io/api/fluxible-context.html#executeaction-action-payload-done-), `executeAction` method has optional callback and returns `executeActionPromise`.

But `this.context.executeAction` seems to return `undefined` and `done` callback is ignored, I fix it.

```jsx
const Component = React.createClass({
  contextTypes: {
    getStore: React.PropTypes.func.isRequired,
    executeAction: React.PropTypes.func.isRequired
  },
  componentDidMount() {
    this.context.executeAction(SomeAction, {}, done); // returns undefined, done won't called.
  },
  render() {
    return <div />;
  }
});

module.exports = provideContext(connectToStores(Component, [SomeStore], (stores) => {
  return stores.SomeStore.getState();
}));
```

### DIFF
--- a/lib/FluxibleContext.js
+++ b/lib/FluxibleContext.js
@@ -203,8 +203,8 @@ FluxContext.prototype.getComponentContext = function getComponentContext() {
     var componentContext = {
         getStore: this._dispatcher.getStore.bind(this._dispatcher),
         // Prevents components from directly handling the callback for an action
-        executeAction: function componentExecuteAction(action, payload) {
-            self.executeAction(action, payload)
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

mridgway:
Hi @geta6, the component context `executeAction` does not return a promise or allow a callback so that we preserve Flux's rule of actions being fire-and-forget (from components). You're not the first to create a PR for this, so it's clear we need to document this reasoning better.

geta6:
I see, thanks.

PR #136, opened 2015-04-17
[Resolves #133] Warn when a callback is passed into executeAction from component

### LINKED ISSUE #133: Warn if users pass in callback to executeAction from a component
I know we documented this but some users skip docs and just learn by coding. It would be nice to warn with a link to the docs like how react does it for certain wqrnings.

Vijar:
I actually wouldn't even be opposed to throwing an error.

@mridgway @redonkulus

### DIFF
--- a/docs/api/Components.md
+++ b/docs/api/Components.md
@@ -9,7 +9,7 @@ The component context receives limited access to the [FluxibleContext](FluxibleC
  * `executeAction(action, payload)`
  * `getStore(storeConstructor)`
 
-It's important to note that `executeAction` does not allow passing a callback from the component. This enforces that the actions are fire-and-forget and that state changes should only be handled through the Flux flow. You may however provide an app level `componentActionHandler` function when instantiating Fluxible. This allows you to handle errors (at a high level) spawning from components firing actions.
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

redonkulus:
Wonder if we should only warn in dev...

Vijar:
Sure, but so many people are misusing this that I think we should take every opportunity to warn.

This isn't just a deprecation warning, it is a common improper usage of the API. Passing in that callback is just as wrong in prod as in dev.

redonkulus:
True, I was just considering spamming people. Which hopefully they would see before they deploy.

mridgway:
I like the idea of throwing as well, but we should wait til next minor version IMO.

cesarandreu:
+1 for waiting for next minor version, wouldn't be nice for people's code to start breaking. :D

I prefer a dev-only warning.

Maybe `if (done)` should be changed to `if (done !== undefined)` (or `if (done !== void 0)` if paranoid), if you want to catch all cases in which people pass in anything, or `if (typeof done === 'function')` if you only want to catch when a function is passed. Checking `if (done)` is a weird in-between.

PR #142, opened 2015-04-22
Stopped FluxibleComponent from double rendering

antonklava:
Fixed case where wrapping component with fluxible would cause component to double render its childen prop.

The test would initially render
```html
<div class="Component" data-reactid=".0"><div class="Component" data-reactid=".0.0">Some child</div></div>
```

This is however at the cost of not passing the props from FluxibleComponent to the component. Perhaps omitting `this.props.children` is a better solution.

### DIFF
--- a/addons/FluxibleComponent.js
+++ b/addons/FluxibleComponent.js
@@ -27,7 +27,7 @@ var FluxibleComponent = React.createClass({
     },
 
     render: function () {
-        return React.addons.cloneWithProps(this.props.children, this.props);
+        return React.addons.cloneWithProps(this.props.children);
     }
 });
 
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

Vijar:
Looks like the `children` key of `this.props` was being passed into the child which caused the wrapped component to have itself as a child component. Not necessarily a double render, but rather unintended nesting of components.

Since `FluxibleComponent` only sets `childContext`, I don't think there is any reason to pass props into the child component it clones so I like this method. 

I'm not sure about the test. @mridgway thoughts?

mridgway:
Whoa, that's really odd. I wonder if that's a bug that should be reported to React.

antonklava:
I'm not sure it's worth reporting to React since, as I see it, the issue is referencing children through `this.props.children` instead of `this.children` or something similar which i suppose was a thought out decision.

I agree that the test may not be the prettiest – is there something I can do there? Otherwise feel free to close/merge as you see fit.

mridgway:
Test looks fine to me :+1:

mridgway:
Thanks!

PR #147, opened 2015-04-27
Add decorator support for provideContext and connectToStores

redonkulus:
@mridgway @jedireza 

Adds support for [ES7 decorators](https://github.com/wycats/javascript-decorators). Basically changes the interface of `connectToStore` and `provideContext` from passing in the React component as the first argument, to returning a function that accepts the React component as an argument.

Before:

```js
class FooComponent extends React.Component {
  render() {
    return <div/>;
  }
}

const ConnectedComponent = connectToStores(FooComponent, 'test');
```

After (with decorator):

```js
@connectToStores('test')
class ConnectedComponent extends React.Component {
  render() {
    return <div/>;
  }
}
```

After (without decorator):

```js
class FooComponent extends React.Component {
  render() {
    return <div/>;
  }
}

const ConnectedComponent = connectToStores('test')(FooComponent);
```

This would break backwards compatibility. I could make it BC friendly by checking the arguments passed in, but that's a little kludgy. 

If good, I can update the docs.


yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

jedireza:
This is an interesting approach. I like the syntax using the decorator. Without the decorator the syntax is less appealing imo because of the bumping parents `connectToStores('test')(FooComponent)`. We could create a new variable out of:

```js
// without decorator

class FooComponent extends React.Component {
  render() {
    return <div/>;
  }
}

const myConnectToStores = connectToStores('test');
const ConnectedComponent = myConnectToStores(FooComponent);
```

If it was BC friendly that would be nice but also more strain to support. I guess another thing we could ask is how much experimental syntax do we feel comfortable using. [Decorators are currently in proposal status](https://babeljs.io/docs/usage/experimental/).

redonkulus:
@mridgway @jedireza 

I refactored this to maintain backwards-compatibility. I had to use `stage0` of babel, which is considered experimental because I needed `static` support for classes. Otherwise, defining things like `contextTypes` or `propTypes` would not be recognized by the decorator pattern.

addons/provideContext.js
```diff
@@ -46,4 +39,39 @@ module.exports = function provideContext(Component, customContextTypes) {
...
+ *   }
+ *
+ * @method provideContext
+ * @param {React.Component} [Component] component to wrap
+ * @param {object} customContextTypes Custom contextTypes to add
+ * @returns {React.Component} or {Function} if using decorator pattern
+ */
+module.exports = function provideContext(Component, customContextTypes) {
+    // support decorator pattern
+    if (arguments.length === 0 || !arguments[0].displayName) {
```
redonkulus: I'm not sure if this is the best way to test if a React component is passed in. I know JSX automatically creates the `displayName` property, but not sure about non-JSX components. If there is a better way, let me know.

mridgway: Yeah, `displayName` may not always be set when not using JSX. The only other thing I can think of is duck-typing by checking for a `render` method on the prototype.

redonkulus: @mridgway another approach could be to just check if a function is passed in: `if (arguments.length === 0 || typeof arguments[0] !== 'function')`...

mridgway: True. That should be fine.

redonkulus: @mridgway updated

PR #153, opened 2015-05-07
this should be isomorphic right?

jedireza:


yahoocla:
CLA is valid!

mridgway:
Services themselves are polymorphic since it's the same interface to two different implementations. The interface is "isomorphic". So technically, it's correct the way it is.

PR #155, opened 2015-05-11
Making promise example less verbose and reducing duplication

DelvarWorld:
Maybe I'm too used to promises but if you can chain a .then to a method it seems obvious to me it's already returning a promise.


yahoocla:

yahoocla:
CLA is valid!

mridgway:
:+1: Thanks!

PR #156, opened 2015-05-11
Avoid calling setState when component is being unmounted

mridgway:
This avoids an issue where you have a parent and a child both listening to the same store, where parent could re-render in the middle of handling the change event causing the child to be unmounted. When this happens, the child's call to `setState` will cause a warning.

It's important to note here that the `isMounted()` may be removed in the future: https://facebook.github.io/react/docs/component-api.html#ismounted. If this happens we will either need to implement our own flag (using `componentWillUnmount`) or hope that an alternative is found. See https://github.com/facebook/react/issues/2787 for discussion on this warning and whether it's needed or not. 

Also fixes https://github.com/yahoo/fluxible-router/issues/18

### DIFF
--- a/addons/connectToStores.js
+++ b/addons/connectToStores.js
@@ -58,7 +58,9 @@ module.exports = function connectToStores(Component, stores, getStateFromStores)
             return state;
         },
         _onStoreChange: function onStoreChange() {
-            this.setState(this.getStateFromStores());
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

jedireza:
:+1:

geekyme:
Thanks!

PR #157, opened 2015-05-12
Fixed race condition

benjamindulau:
I'm not sure about this one but when a component is owned by another one and both listen to the same store changes; if the child component in unmounted, `_onStoreChange` is called right after `componentWillUnmount` and so a `setState` call is initiated on it, and it shouldn't.

I just added a test to make sure the component if mounted before trying to call `setState`.

That fixes my use-case but maybe that worth looking more into it.


yahoocla:
CLA is valid!

yahoocla:

benjamindulau:
Oops, I just saw that this is a duplicate of #156, so closing... ;-)

PR #162, opened 2015-05-14
Support for IE8 added when using promise.catch(func).

lucianlature:
## Usage in IE<9

`catch` is a reserved word in IE<9, meaning `promise.catch(func)` throws a syntax error. There are some workarounds specifically for IE8.

However, please remember that such techniques are already provided by most common minifiers, making the resulting code safe for old browsers and production, but when using `browserify` as bundler and `uglify` as minifer, the issues on IE8 are still present.

So I've replaced `.catch` with `.then`

```js
promise.then(undefined, function(err) {
  // ...
});
```

### DIFF
--- a/lib/FluxibleContext.js
+++ b/lib/FluxibleContext.js
@@ -239,10 +239,10 @@ FluxContext.prototype.getComponentContext = function getComponentContext() {
         // Prevents components from directly handling the callback for an action
         executeAction: function componentExecuteAction(action, payload) {
             self.executeAction(action, payload)
-            .catch(function actionHandlerWrapper(err) {
+            .then(undefined, function actionHandlerWrapper(err) {
                 return self.executeAction(self._app._componentActionHandler, { err: err });
             })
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

donatasbacius:
Would be awesome if someone could merge and bump the version asap. I'm in the same position atm  with an app that'll have to go into production soon. 

Could have also used ``promise['catch']() `` but it's a matter of preference.

pkiddie:
+1

redonkulus:
@lucianlature I'd prefer `promise['catch']` as @donatasbacius suggested. We have used this in the past for other libs.

mridgway:
:+1:

mridgway:
This has been released in 0.4.8. Thank you!

donatasbacius:
Awesome, thanks @mridgway and @lucianlature !

PR #164, opened 2015-05-18
[resolves #143] Fix precedence for store name

### LINKED ISSUE #143: connectToStores: getStateFromStores - 'stores' contents get mangled by Uglify
Consider this code:

```javascript
// FooStore.js
class FooStore extends BaseStore {
    // ...
}
FooStore.storeName = 'FooStore';

// BarStore.js
let BarStore = createStore({
    storeName: 'BarStore'
});
export default BarStore;

// MyStore.js
var MyStore = function (dispatcherInterface) {
    BaseStore.apply(this, arguments);
};
inherits(MyStore, BaseStore);
MyStore.storeName = 'MyStore';
export default MyStore;

// Component.jsx
Component = connectToStores(Component, [FooStore, BarStore, MyStore], function (stores, props) {
    return {
        foo: stores.FooStore.getFoo(), // stores.FooStore = undefined
        bar: stores.BarStore.getBar(), // stores.BarStore = ok
        thing: stores.MyStore.getThing() // stores.MyStore = ok
    };
});
```

Without Uglification (default settings), `stores.FooStore`, `stores.BarStore` and 'stores.MyStore' work fine. After Uglifying with default settings, `stores.FooStore` will be undefined, `stores.BarStore` and `stores.MyStore` will work fine.

Stores extending BaseStore with ES6 classes using Babel get turned into named functions. Extending BaseStore like in the MyStore example creates an anonymous function, and  createStore seems to do the same behind the scenes. 

Now in connectToStores, this happens:

```javascript
stores.forEach(function (store) {
    var storeName = store.name || store.storeName || store;
    storeInstances[storeName] = this.context.getStore(store);
}, this);
return getStateFromStores(storeInstances, this.props);
```

Stores made with `createStore` or the 'manual' way to extend `BaseStore` will be anonymous functions, so for them their `storeName` will be used. For stores made with `class Store extends BaseStore`, `store.name` will exist and be set to the class name. However, Uglify mangles the name by default and the store will be available in `storeInstances` with the mangled name only.

I think taking `store.name` out of the `storeName` is enough to fix this. `Function.name` is unreliable after Uglify in any case. Maybe add a check to the store registration when `process.env.NODE_ENV !== 'production'` to see if the `storeName` property was set?

mridgway:
This will fix the `connectToStores` functionality in minified environments as long as users specify a `storeName` static property.

A separate PR should be done in dispatchr to warn users if they rely on the `store.name` property only.

### DIFF
--- a/addons/connectToStores.js
+++ b/addons/connectToStores.js
@@ -43,7 +43,7 @@ module.exports = function connectToStores(Component, stores, getStateFromStores)
             if ('function' === typeof getStateFromStores) {
                 var storeInstances = {};
                 stores.forEach(function (store) {
-                    var storeName = store.name || store.storeName || store;
+                    var storeName = store.storeName || store.name || store;
                     storeInstances[storeName] = this.context.getStore(store);
                 }, this);
                 return getStateFromStores(storeInstances, this.props);


yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #174, opened 2015-05-27
Update executeAction in MockActionContext to return the action.

ross-pfahler:
It now matches the real executeAction code. This is useful for
testing Promises.


yahoocla:
CLA is valid!

mridgway:
Thanks!

PR #175, opened 2015-05-27
Set fallback value for contextPlugin

cesarandreu:
This PR adds a fallback value when contextPlugin is created, so it'll default to an empty object.

The use-case for this is that I'm toying around with the idea of moving my bootstrap logic into a plugin, since it's pretty similar in both of my server and client entry points. When `plugContext`gets called, it'll add `bootstrap` to the context, and likely tweaking some values based on the options it receives.

If you think it's better to keep current behavior I'm fine with that, since I can just return an empty object from my plugin, and I'm not even sure if it's *actually* a good idea to do this, I'm just sorta experimenting.

As an aside, this is my first PR from 35,000 feet in the air :D.

### DIFF
--- a/lib/Fluxible.js
+++ b/lib/Fluxible.js
@@ -55,7 +55,7 @@ Fluxible.prototype.createContext = function createContext(options) {
     // Plug context with app plugins that implement plugContext method
     this._plugins.forEach(function eachPlugin(plugin) {
         if (plugin.plugContext) {
-            var contextPlugin = plugin.plugContext(options, context, self);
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

cesarandreu:
Basically, this lets you plug just the context.

mridgway:
Thanks! Your PRs from 35,000 feet are just as good as others :wink:

cesarandreu:
Thanks :D

PR #176, opened 2015-05-27
[resolves #172, resolves #173] Hoist non-react statics in higher order components

### LINKED ISSUE #172: connectToStores destroys component's statics.
Hello,
I just found that after a component is wrapped by using connectToStores addon:
https://github.com/yahoo/fluxible/blob/master/addons/connectToStores.js
it's missing the statics declared, i.e. the following code will produce the error

**Uncaught TypeError: MyComponent.customMethod is not a function**

```javascript
var connectToStores = require('fluxible/addons').connectToStores;
var MyStore = require('./MyStore');
var MyComponent = React.createClass({
  statics: {
    customMethod: function(foo) {
      return foo === 'bar';
    }
  },
  render: function() {
  }
});
MyComponent= connectToStores(MyComponent, [MyStore], function(stores){
    return stores.MyStore.getState();
});
MyComponent.customMethod('bar');  
```
Is it a bug or feature? :)

### LINKED ISSUE #173: Higher-order Components should hoist statics

mridgway:
This will allow higher order components to be treated as if they are the same component that they wrap. This util will also be used in `fluxible-router` for the same purpose.


yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

redonkulus:
:+1:

cesarandreu:
@mridgway I'd vote for publishing `hoistNonReactStatics.js` as a separate module. It seems like something that would be convenient to other libraries.

mridgway:
Sure we can do that then. Not sure what to name it though... `hoist-non-react-statics`?

cesarandreu:
Yeah, that seems descriptive. I figure something like this would probably get picked up mostly by word of mouth.

mridgway:
Updated to use https://github.com/mridgway/hoist-non-react-statics

cesarandreu:
:+1:

PR #177, opened 2015-05-28
Async server-side rendering: Add `storesLoaded` method to FluxibleContext; server.js can use this method to allow all actions to complete before rendering.

JoshuaGross:
Hey all, there seems to be some confusion online as to how to make React fully isomorphic (including fetching any data necessary), and this is my initial attempt. I'm trying to figure out how to fetch data asynchronously and make sure that HTML is fully inflated before rendering server-side using Fluxible. I was running into an issue where actions were in-flight but were not yet in `currentAction` of the dispatcher context; because of this race condition, pages were being rendered without any actual data. I added some simple book-keeping to FluxibleContext and an optional method that users can hook into to get a callback once all data is loaded. 

In my `server.js`, my `server.use` now contains this block:

```
        context.getActionContext().storesLoaded(function () {
            debug('Exposing context state');
            const exposed = 'window.App=' + serialize(app.dehydrate(context)) + ';';

            debug('Rendering Application component into html');
            const html = React.renderToStaticMarkup(htmlComponent({
                assets,
                context: context.getComponentContext(),
                state: exposed,
                markup: React.renderToString(context.createElement())
            }));

            debug('Sending markup');
            res.type('html');
            res.write('<!DOCTYPE html>' + html);
            res.end();
        });
```

I'm fairly new to React so I'm somewhat concerned I need this because of some anti-pattern I'm using, but Fluxible's documentation and various articles online didn't help much. So, I'd love feedback. Thanks!

### DIFF
--- a/lib/FluxibleContext.js
+++ b/lib/FluxibleContext.js
@@ -35,6 +35,9 @@ function FluxContext(options) {
     this._actionContext = null;
     this._componentContext = null;
     this._storeContext = null;
+
+    // Book-keeping for actions
+    this.executingActions = 0;
 }
 
 // Provied a way to override Promise only for FluxContext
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

mridgway:
@JoshuaGross You shouldn't really need this. When you call `context.getActionContext().executeAction(fooAction, payload)` you will receive a promise that will only be fulfilled when that action and its dispatches have completed. If that's not the case, you are likely returning too early from your action or not returning a Promise when performing asynchronous operations.

JoshuaGross:
@mridgway I'm trying to: (1) avoid explicitly loading any data in `server.js`, which I would like to remain agnostic of any data-usage patterns. The chatroom example seems to assume that we always want to load messages on every page, so it calls `showMessages` which I want to avoid. (2) I want to avoid duplicating code on server.js and client.js for obvious reasons. 

Instead, what I'm doing is that a component (when it's created) will dispatch an event to load the data it needs. Is there a better way?

PR #183, opened 2015-06-05
Moved FluxibleContext.createElement to addons as createElementWithContext

mridgway:
This is the first step to making Fluxible not depend on React directly. This is the only method within core that uses React. By moving this to addons, we can split addons to a separate package and core will work independently with any framework (React/React Native/etc.).


yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #195, opened 2015-06-17
Allow rehydration with empty object

mridgway:
Safeguards rehydration from throwing undefined errors.


yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #198, opened 2015-06-20
[resolves #124] Pass full context to connectToStore getState function…

mridgway:
… to allow more flexibility; Also allow adding additional context types

See #124 for discussion.


redonkulus:
:+1:

PR #202, opened 2015-06-24
Use fluxible-addons-react temporarily for backwards compatibility; Wa…

mridgway:
…rn on usage of addons directly from fluxible

This removes Fluxible's dependency on React. Fluxible will still have a dependency on fluxible-addons-react for one minor version with warnings if the addons are used directly from `fluxible/addons`. This will allow us to completely remove the React dependency from the dep chain in the next version.

The plan at this point is to 1.0 Fluxible after this upcoming minor version.


redonkulus:
:+1:  after fluxible.io is updated.

mridgway:
Just found some things to fix:

 - [ ] npm 1.x fails on peer dependency
 - [ ] Add warnings for `fluxible/addons` usage.

addons/provideContext.js
```diff
@@ -1,80 +1,6 @@
...
-
-        render: function () {
-            return React.createElement(Component, this.props);
-        }
-    });
-
-    hoistNonReactStatics(ContextProvider, Component);
-
-    return ContextProvider;
+if (process.env.NODE_ENV === 'production') {
```
Vijar: Shouldn't we warn in dev?

mridgway: Yes. Fixing.

Vijar:
Build will fail because `fluxible-addons-react` depends on this PR being published...so lets :shipit: .

PR #211, opened 2015-07-07
remove unused require

escalona:


tleunen:
You should remove it from the `package.json` as well

escalona:
@tleunen It's still used here: https://github.com/yahoo/fluxible/blob/e6ead51382f366e3ac5784269ada726786e8f47f/addons/index.js#L13

tleunen:
you're right. Hopefully we'll be able to fully remove it

redonkulus:
:+1:

PR #212, opened 2015-07-08
Add quad-blog to reference applications

cesarandreu:
My blog :D.


cesarandreu:
I guess some interesting highlights are:

* Immutable.js stores
* Shows how you'd do all the `<head>` tags
* Webpack example of development and production environment for both client-side and server-side code, without sacrificing webpack's ability to import assets!
* No gulp / grunt needed, cuz webpack ftw

yahoocla:
CLA is valid!

mridgway:
Nice!

PR #223, opened 2015-07-12
Update BaseStore.md

leebenson:
Incorrectly referenced `fluxible-addons-react/BaseStore`, which doesn't exist


mridgway:
Thanks!

PR #226, opened 2015-07-14
add action name info error object while execute action

kaesonho:
1. this might be useful when we want to integrate `componentActionHandler`, as long as payload.errorInfo is defined, append `info` to error object for users to decide how to handle the error.

2. `action.displayName || action.name` might not work with IE, add some code to be compatible with IE

@mridgway @redonkulus @lingyan

### DIFF
--- a/lib/FluxibleContext.js
+++ b/lib/FluxibleContext.js
@@ -99,12 +99,27 @@ FluxContext.prototype.plug = function (plugin) {
 };
 
 /**
+ * Get action name, compatible with all browsers
+ * @param {Function} action
+ * @return {String} action display name
+ */
+function getActionName (action) {
+    var actionName = action.displayName || action.name;
+    if (!actionName) {
+        var match = action.toString().match(/^function\s*([^\s(]+)/);
+        actionName = match && match[1];
+    }
+    return actionName;
+};
+
+/**
  * Executes an action, and calls either resolve or reject based on the callback result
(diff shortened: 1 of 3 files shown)


lib/FluxibleContext.js
```diff
@@ -99,12 +99,27 @@ FluxContext.prototype.plug = function (plugin) {
...
 
 /**
+ * Get action name, compatible with all browsers
+ * @param {Function} action
+ * @return {String} action display name
+ */
+function getActionName (action) {
+    var actionName = action.displayName || action.name;
+    if (!actionName) {
+        var match = action.toString().match(/^function\s*([^\s(]+)/);
```
mridgway: I would think this would only contain the name if `action.name` is populated.

cesarandreu: I thought `Function.name` was standard, but then I looked it up in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name)... Stupid IE. (╯°□°）╯︵ ┻━┻

kaesonho: I think it's fine since actionName here is only debug usage, 
I generated this for some other purpose which is not valid anymore, just keep the code

kaesonho:
This PR will be kind of specific, we've got some other approach to do the same thing.

let me close this :)

PR #238, opened 2015-08-04
Added Contactor to reference applications

localnerve:
A data driven example with source.  

Source: https://github.com/localnerve/flux-react-example. 
Live: http://flux-react-example.herokuapp.com


mridgway:
Thanks!

yahoocla:
CLA is valid!

PR #241, opened 2015-08-10
Adds a safe global check

pablolmiranda:
Checks if the global object exists before try to access the Promise object in it.
This creates a `Cannot read property 'Promise' of undefined` TypeError when trying to load browserify and webpack modules at the same time.
@redonkulus


redonkulus:
:+1:

yahoocla:
CLA is valid!

PR #254, opened 2015-09-08
[docs] prebind this.onClick in constructor of ES6 Class

### LINKED ISSUE #253: Unable to get context from the onclick event
I want to bind a click event just like the actions example.

http://fluxible.io/api/actions.html

So i do it like this:

    function testAction(actionContext, payload, done) {
        setTimeout(function () { // simulate async
            actionContext.dispatch('MY_ACTION', payload);
            done();
        }, 10);
    }

    class Home extends React.Component {
	static contextTypes = {
            executeAction: React.PropTypes.func.isRequired,
            getStore: React.PropTypes.func.isRequired
        }

        onClick() {
    	    console.log(this)
    	    this.context.executeAction(testAction, {});
        }

        render() {
    	    var list = this.context.getStore(TestStore).getTestData()

    	    var listHtml = list.map(function(i){
    		    return <div>{i}</div>
    	    })

            return (
                <div>
                    <h2>Home</h2>
                    <p>Welcome to the site!</p>
                    <div>{listHtml}</div>
                    <button onClick={this.onClick}>click me</button>
                </div>
            );
         }
 
    }

But it doesn't work because the variable 'this' is undefined. Ather searching on stackoverflow, i figure it out by the following code:

    <button onClick={this.onClick.bind(this)}>click me</button>

Does the doc has something wrong ? 
Do I get the wrong way to use component ?

koulmomo:
ES6 `class` syntax for components does not have autobinding lie with `React.createClass`

explicitly prebinding method in the constructor as shown [here](https://github.com/facebook/react/blob/master/docs/_posts/2015-01-27-react-v0.13.0-beta-1.md#autobinding)

resolves #253

### DIFF
--- a/docs/api/Actions.md
+++ b/docs/api/Actions.md
@@ -64,6 +64,7 @@ import myAction from './myAction';
 class MyComponent extends React.Component {
     constructor (props) {
         super(props);
+        this.onClick = this.onClick.bind(this);
     }
     onClick (e) {
         this.context.executeAction(myAction, {});


redonkulus:
:+1:

yahoocla:
CLA is valid!

redonkulus:
Thanks @koulmomo

Issue #208, opened 2015-06-30
Remove flux-router-component from package


redonkulus:
Since this is deprecated, we should remove it. There are tests and docs that refer to it still.

mridgway:
Closed via #257.

PR #257, opened 2015-09-10
Remove fluxible-addons-react from dependencies

mridgway:
Resolves https://github.com/yahoo/fluxible/issues/256

Also removes deprecated APIs and tests that have been moved to `fluxible-addons-react`.


mridgway:
Now also resolves https://github.com/yahoo/fluxible/issues/208.

yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #258, opened 2015-09-10
Remove pre-ES2015 dependencies

mridgway:


redonkulus:
we should probably mention this in the readme now, no? if you are on node 10 and want to run unit tests, what do you do?

yahoocla:
CLA is valid!

mridgway:
Unit tests use babel which automatically adds Promise.

redonkulus:
Ah good point. I still think the readme should be updated so people are not bitten by this.

mridgway:
Updated with a note in the README.

redonkulus:
:+1:

PR #261, opened 2015-09-11
Add a note about provideContext to the plugins doc

whittle:
When authoring plugins, people may not realize that there is an additional step to make plugin functionality available to all components. This patch adds a nota bene linking to the appropriate doc.


bradbarrow:
:+1: This was a gotcha for me thanks @whittle

yahoocla:

yahoocla:
CLA is valid!

mridgway:
Nice. I will look into making this documentation a little clearer as well.

PR #262, opened 2015-09-14
Updating docs; Restructuring to remove outdated articles and link to libraries instead

### LINKED ISSUE #91: Document `componentActionHandler` with an example of how to use it
Possibly reference https://github.com/yahoo/fluxible-plugin-routr/issues/16. This could be a guide for the docs.

/cc @jedireza

mridgway:
Resolves #91

Some of these "guides" were just walking through setting up other libraries like fluxible-router and fetchr plugin. It think it will be better to just link to the corresponding library's docs since they will be better kept up-to-date. For now I removed the routing guide and moved the "bringing-flux-to-the-server" to a blog post since it is not updated. I'm considering removing the fetchr guide as well since it seems to be the same steps as https://github.com/yahoo/fluxible-plugin-fetchr#usage.


docs/community/articles.md
```diff
@@ -0,0 +1,6 @@
+# Articles
```
redonkulus: add http://dev.alexishevia.com/2014/11/isomorphic-react-flux-using-yahoos.html

mridgway: Added a bunch more

redonkulus:
:+1:

yahoocla:
CLA is valid!

PR #263, opened 2015-09-15
Action tracing

Vijar:
@redonkulus @mridgway


tests/unit/lib/FluxibleContext.js
```diff
@@ -148,13 +149,84 @@ describe('FluxibleContext', function () {
...
+                    expect(actionCalls[2].payload).to.equal(payload);
+                    expect(actionCalls[3].context).to.contain.keys(Object.keys(actionContext));
+                    expect(actionCalls[3].context).to.contain.keys(['id','stack']);
+                    expect(actionCalls[3].context.id).to.equal('THIS_IS_UNIQUE');
+                    expect(actionCalls[3].context.stack.join('.')).to.equal('Action.One.Three');
+                    expect(actionCalls[3].payload).to.equal(payload);
+                    done();
+                };
+                actionContext = context.getActionContext({
+                    id: 'THIS_IS_UNIQUE'
```
Vijar: I'm not sure if we should provide the id in the getActionContext call or the executeAction call.

tests/unit/lib/FluxibleContext.js
```diff
@@ -441,6 +513,81 @@ describe('FluxibleContext', function () {
...
+                        payload: payload
+                    });
+                    callback();
+                };
+                actionThree.displayName = 'Three';
+                var payload = {};
+                // delete componentContext cache to regenerate it with id this time
+                delete context._componentContext
+                componentContext = context.getComponentContext({
+                    id: 'THIS_IS_UNIQUE'
```
Vijar: `getComponentContext` is really only called once on the client so maybe it makes sense to pass in a unique id when calling `executeAction` instead.

lib/FluxibleContext.js
```diff
@@ -190,46 +197,81 @@ FluxContext.prototype._initializeDispatcher = function initializeDispatcher() {
...
+                /*
+                 * Pass the current stack along.
+                 * If this action calls more actions,
+                 * we keep modifying and passing the stack along.
+                 */
+                var actionName = action.displayName || action.name;
+                // we have to clone the stack, can't just push to it
+                // since one action can execute multiple actions.
+                var nextStack = this.stack.concat([actionName]);
+                return self.executeAction(action, payload, callback, {
```
Vijar: We create a new `meta` parameter which holds things like the action stack trace and the id. This `meta` object is just propagated to all the actions that are executed by the first action.

lib/FluxibleContext.js
```diff
@@ -190,46 +197,81 @@ FluxContext.prototype._initializeDispatcher = function initializeDispatcher() {
...
+            if (actionContextPlugin) {
+                actionContextPlugin(actionContext, self, self._app);
+            }
+        });
+
+        self._actionContext = actionContext;
+    }
 
-    return self._actionContext;
+    var nextStack = (meta && meta.nextStack) || ['Action'];
```
Vijar: The first action will start with `Action` as the start of the stack trace. Usually this would be `navigateAction`

lib/FluxibleContext.js
```diff
@@ -190,46 +197,81 @@ FluxContext.prototype._initializeDispatcher = function initializeDispatcher() {
...
+     * a shallow clone with a new stack every time getActionContext is called
+     */
+    return objectAssign({}, self._actionContext, {
+        stack: nextStack,
+        id: (meta && meta.id) || generateUUID()
+    });
 };
 
+var tempThing = 0;
+function generateUUID () {
```
Vijar: What is a good way to generate UUIDs if the user doesn't provide an id?

lib/FluxibleContext.js
```diff
@@ -190,46 +197,81 @@ FluxContext.prototype._initializeDispatcher = function initializeDispatcher() {
...
+     * of the actionContext interface.
+     * You can access this in your actions with `context.stack`.
+     * Use the `displayName` property on your actions for better
+     * action tracing when your code gets minified in prod.
+     * One action can execute multiple actions, so we need to create
+     * a shallow clone with a new stack every time getActionContext is called
+     */
+    return objectAssign({}, self._actionContext, {
+        stack: nextStack,
+        id: (meta && meta.id) || generateUUID()
```
Vijar: The first action executed can set this id. If not, one is generated.

lib/FluxibleContext.js
redonkulus: you could also do something like: 
```
function guid() {
    var hrtime = process.hrtime();
    return hrtime[0] * 1e9 + hrtime[1];
}
```
which uses nanosecond and should be unique enough. This way you can remove the temporary variable.

lib/FluxibleContext.js
```diff
@@ -134,9 +140,10 @@ function executeActionInternal(context, action, payload, resolve, reject) {
  *  and done as parameters
  * @param {Object} payload The action payload
  * @param {Function} [done] Method to be called once action execution has completed
+ * @param {Object} meta Metadata to keep track of the current flow of executing actions
  * @return {Promise} executeActionPromise Resolved with action result or rejected with action error
  */
-FluxContext.prototype.executeAction = function executeAction(action, payload, done) {
+FluxContext.prototype.executeAction = function executeAction(action, payload, done, meta) {
```
redonkulus: could we maintain BC by adding meta to the payload as a property? feels odd to not have done last.

Vijar: Yeah, I guess so. But now `meta` will be a reserved property that we override on the payload...that seems more breaking than adding a new property right?

redonkulus: is meta meant to be used by users as well or just for fluxible to use?

tests/unit/lib/FluxibleContext.js
redonkulus: will the user ever pass it in manually to executeAction?

lib/FluxibleContext.js
mridgway: This is how React generates random root index IDs on the server: https://github.com/facebook/react/blob/master/src/renderers/dom/server/ServerReactRootIndex.js. It's important that you don't use a simple increment since it will eventually overflow.

On the client they use the same method as you did: https://github.com/facebook/react/blob/master/src/renderers/dom/client/ClientReactRootIndex.js

lib/FluxibleContext.js
```diff
@@ -173,46 +180,81 @@ FluxContext.prototype._initializeDispatcher = function initializeDispatcher() {
...
  * Returns the context for action controllers
  * @method getActionContext
+ * @param {Object} meta Metadata to keep track of the current flow of executing actions
  * @return {Object} Action context information
  */
-FluxContext.prototype.getActionContext = function getActionContext() {
-    if (this._actionContext) {
-        return this._actionContext;
-    }
+FluxContext.prototype.getActionContext = function getActionContext(meta) {
```
mridgway: Remove this parameter.

lib/FluxibleContext.js
```diff
@@ -173,46 +180,81 @@ FluxContext.prototype._initializeDispatcher = function initializeDispatcher() {
...
+    return ++tempThing;
+}
+
 /**
  * Returns the context for action controllers
  * @method getComponentContext
  * @return {Object} Component context information
  */
-FluxContext.prototype.getComponentContext = function getComponentContext() {
+FluxContext.prototype.getComponentContext = function getComponentContext(meta) {
```
mridgway: Remove this parameter.

lib/FluxibleContext.js
mridgway: We can internalize this function and only expose a three parameter function publicly. All publicly facing `executeAction`s will now go through the new internal executeAction instead of the public facing one.

lib/FluxibleContext.js
mridgway: Let's just always generate this. I don't see value in allowing it to be passed in.

lib/FluxibleContext.js
mridgway: For actions it would be rare to reach max int, but it may be possible. Perhaps it's safe to always use the randomized one from React but make sure it supports all browsers.

tests/unit/lib/FluxibleContext.js
mridgway: For testing since we won't allow it to be passed in, you can mock out the generatUUID method.

Vijar: We decided to always autogenerate this

lib/FluxibleContext.js
Vijar: Just for fluxible. I'm refactoring to still maintain the three params here.

lib/FluxibleContext.js
Vijar: After discussing offline, we decided while we can remove this parameter and move the objectAssign logic into `executeActionInternal` this is probably the best place for in since the `actionContext.executeAction` function's `this` depends on the `objectAssign` being done successfully.

Vijar:
@redonkulus @mridgway updated.

yahoocla:
CLA is valid!

mridgway:
:+1:

lib/FluxibleContext.js
```diff
@@ -83,12 +84,17 @@ FluxContext.prototype.plug = function (plugin) {
...
  * @param {Function} resolve function to call on success
  * @param {Function} reject function to call on failure
  * @private
  */
-function executeActionInternal(context, action, payload, resolve, reject) {
-    var syncResult = action(context.getActionContext(), payload, function (err, result) {
+function executeActionInternal(context, action, payload, meta, resolve, reject) {
+    var actionContext = context.getActionContext(meta);
+    var displayName = action.displayName || action.name;
+    debug('Executing action ' + displayName + ' with trace', actionContext.stack.join('.'));
```
redonkulus: minor: only thing i dont like is that this join happens even if debug isn't used

redonkulus:
can we create a doc for this and how users can use it?

Vijar:
LGTM. Just need docs now.

Vijar:
:ship:

redonkulus:
This looks fine, I would just really like to have docs too before we forget and never do it.

mridgway:
I am contemplating renaming this `context.id` to `context.rootId` to make it more clear that it is not changing for every action context. In the future, it may be useful to have a separate `id` that is unique for every action.

Vijar:
I'm fine with that. Should we come up with a better name? 

What do we call a cascade of actions? A trasaction?

PR #269, opened 2015-09-23
Correctly bind executeAction to newactionContext

Vijar:
@mridgway @kaesonho 

`executeAction` is throwing `Cannot read property 'stack' of null` when `context.executeAction` was called via async:
```
async.apply(context.executeAction, myAction, payload)
```

This would also happen if a user did something like this as @mridgway pointed out:
```js
var exec = context.executeAction;
exec(myAction, payload);
```


mridgway:
:+1:

yahoocla:
CLA is valid!

mridgway:
I will backport this one merged.

gfranko:
:+1:

Issue #273, opened 2015-09-28
Monorepo


mridgway:
I'm considering whether we should move many of the official packages under the same git repo rather than having them spread out across the Yahoo org. As an example, you can see [Babel's explanation](https://github.com/babel/babel/blob/development/doc/design/monorepo.md) of why they prefer a single repo for multiple packages rather than many.

I tend to agree, especially when we are doing upgrades for React versions, it can be a pain to coordinate multiple repos rather than being able to do everything at once as part of one repo.

Here is a list of what we (the Fluxible team) currently maintain:

Fluxible specific (more likely to have common repo)
* [x] `fluxible` (#274)
* [x] `fluxible-reducer-store` (#275)
* [x] `generator-fluxible`
* [x] `fluxible.io`
* [x] `flux-examples`
* [x] `fluxible-addons-react`
* [x] `fluxible-router`
* [x] `fluxible-plugin-fetchr`
* [x] `dispatchr`

Standalone (more likely to stay independent)
* `routr` (_internal to fluxible-router but used by others like redux-universal-router_)
* `hoist-non-react-statics` (_generic_)
* `fetchr` (_generic_)

cesarandreu:
+1

Managing lots of repos is hard, if moving it into a single repo helps you iterate on things it sounds like a big win.

mdressman:
+1

no more `npm link` madness

pablolmiranda:
:+1:

akshayp:
:100: A long but good read on the topic from twitter's experience with bringing things into a monorepo http://www.gigamonkeys.com/flowers/

redonkulus:
:+1: would make our lives easier for sure. Like the idea of having the docs and website in one repo, will make fluxible.io easier to parse the docs too.

mridgway:
I'm leaning towards moving `dispatchr` in since it's a core piece of Fluxible. It would still remain uncoupled.

redonkulus:
> I'm leaning towards moving dispatchr in since it's a core piece of Fluxible. It would still remain uncoupled.

:+1:

localnerve:
This is cool, but always reminds me "this is more of a Shelbyville idea".

mridgway:
It is done. There may still be some links that are broken, so if you see any 404s please let us know.

PR #274, opened 2015-10-01
[#273] Restructure for monorepo

mridgway:
Moves fluxible npm package files into `packages/fluxible`. Other libraries will follow same pattern: `packages/<packageName>`.

Added gulp for installing all package dependencies and linking them to the root. This will make it easier for updating all modules at the same time.

Problems to think about:

* There will only be one Travis-CI job, so the tests will run for all packages on every PR/commit. We can use matrix to run each package as a separate job which I'm trying out here: https://github.com/yahoo/fluxible/pull/274/files#diff-354f30a63fb0907d4ad57269548329e3R10

* Code coverage will be for the full repo instead of for each individual package. I'm not sure how this will work with the package matrix mentioned above. We would likely have to run a full suite instead of each package individually for this to work.

* Git tags would overlap, so we'll have to start using the package name as part of the tag. For this we should use a custom `version` command rather than using `npm version`.


Vijar:
:ship:

yahoocla:
CLA is valid!

PR #276, opened 2015-10-08
pin fluxible-addons-react to 0.1 to make sure react dep is met, since…

lingyan:
… fluxible-addons-react@0.2 (the latest on npm) wants react-0.14

This PR is for fluxible@0.5.x

@mridgway


Vijar:
:+1:

PR #281, opened 2015-10-15
Add `fluxible-plugin-actions` to `Actions` lib

gingur:
This plugin allows components to get action creators without knowing file structure, similar to how get store works. troy.rhinehart@gmail.com


mridgway:
Awesome thanks! We may want to add docs to recommend this for use with `fluxible-router` at some point: https://github.com/yahoo/fluxible-router/blob/master/lib/navigateAction.js#L38.

yahoocla:
CLA is valid!

PR #295, opened 2015-10-28
resolve #294, MockActionContext Promise support

### LINKED ISSUE #294: MockAction for use with MockActionContext
Wondering if it is a good idea to add a MockAction (wrapper) for use with MockActionContext to supply Promise fulfillment if 'done' callback is missing, or just update the MockXXXContext utils directly to use a common fulfillment mechanism.

More and more my code is making use of Promises, and I suspect this is growing for everyone.

I rely pretty heavily on the Fluxible Utils, so if these are going to stick around (which I super-hope they do), then they probably should get an update for actions that make use of Promises in one way or another.

Other Ideas?

localnerve:
My idea for #294. (re)Uses the Fluxible callAction idea for MockActionContext. MockComponentContext not needed.

### DIFF
--- a/packages/fluxible/utils/callAction.js
+++ b/packages/fluxible/utils/callAction.js
@@ -1,0 +1,40 @@
+/**
+ * Copyright 2014, Yahoo! Inc.
+ * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
+ */
+/* global Promise */
+'use strict';
+var isPromise = require('is-promise');
+
+/**
+ * Call an action supporting Promise expectations on invocation.
+ *
+ * If done callback supplied, that indicates non-Promise invocation expectation,
+ * otherwise, Promise invocation.
+ */
+function callAction (action, context, payload, done) {
+  if (done) {
+    return action(context, payload, done);
+  }
+
+  return new Promise(function (resolve, reject) {
+    try {
+      var syncResult = action(context, payload, function (err, result) {
+        if (err) {
+          reject(err);
+        } else {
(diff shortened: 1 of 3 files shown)


packages/fluxible/utils/callAction.js
```diff
@@ -0,0 +1,40 @@
...
+var isPromise = require('is-promise');
+
+/**
+ * Call an action supporting Promise expectations on invocation.
+ *
+ * If done callback supplied, that indicates non-Promise invocation expectation,
+ * otherwise, Promise invocation.
+ */
+function callAction (action, context, payload, done) {
+  if (done) {
```
mridgway: I think we don't want to short circuit here. It should always return a `Promise` just like `executeAction` does: https://github.com/yahoo/fluxible/blob/master/packages/fluxible/lib/FluxibleContext.js#L174

Did you see an issue that caused you to add this check?

mridgway:
This looks good! One question above, but otherwise looks great.

yahoocla:
CLA is valid!

packages/fluxible/utils/callAction.js
localnerve: I just wanted to keep it 'as-is' for the caller unless the caller specifically wants a different behavior.

mridgway:
I'll test this out on some of our other test suites (fluxible-router and internal) to verify.

mridgway:
:+1: thanks @localnerve!

mridgway:
Released v1.0.1 with this enhancement.

PR #297, opened 2015-11-01
Add rootId to MockActionContext

### LINKED ISSUE #296: Consider adding the transaction ID to MockActionContext
Hi, I'd like to test a [use case](https://github.com/yahoo/fluxible-router/pull/95) where I use `context.rootId` inside an action. I expect I can write something something like

```js
mockContext = createMockActionContext({ stores: [...] });
console.log(mockContext.rootId); // print a uuid
```

I'd send a PR to change [MockActionContext](https://github.com/yahoo/fluxible/blob/a1dd75ebf2e76456ad2ec903e77104cc70b9ec44/packages/fluxible/utils/MockActionContext.js):

```js
// same function from FluxibleContext.js
var GLOBAL_UUID_MAX = Math.pow(2, 53);
function generateUUID () {
    return Math.ceil(Math.random() * GLOBAL_UUID_MAX);
}
function MockActionContext (dispatcherContext) {
    this.dispatcherContext = dispatcherContext;
    this.executeActionCalls = [];
    this.dispatchCalls = [];
    this.rootId = generateUUID(); 
}
```

... but if right shouldn't we before export that `generateUUID()` somewhere? 

Thanks!

gpbl:
This PR extracts the `generateUUID` function into its own util and adds a `rootId` to `MockActionContext`

### DIFF
--- a/packages/fluxible/lib/FluxibleContext.js
+++ b/packages/fluxible/lib/FluxibleContext.js
@@ -6,6 +6,8 @@
 
 var debug = require('debug')('Fluxible:Context');
 var isPromise = require('is-promise');
+var generateUUID = require('../utils/generateUUID');
+
 var __DEV__ = process.env.NODE_ENV !== 'production';
 require('setimmediate');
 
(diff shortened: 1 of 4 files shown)


mridgway:
:+1:

yahoocla:
CLA is valid!

PR #298, opened 2015-11-04
Checks if the action is a function for utils/callAction.js

pablolmiranda:
Emulates the DEV behavior of the FluxibleContext executeAction since this helper is mainly use for tests.
The executeAction should thrown a error if the action is not a function.


redonkulus:
:+1:

yahoocla:
CLA is valid!

PR #308, opened 2015-11-12
Fix stack tracing for deep stacks

mridgway:
This solves an issue with stack tracing not working on deeply nested `executeAction` calls.

### DIFF
--- a/packages/fluxible/tests/unit/lib/FluxibleContext.js
+++ b/packages/fluxible/tests/unit/lib/FluxibleContext.js
@@ -134,12 +134,20 @@ describe('FluxibleContext', function () {
                         context: context,
                         payload: payload
                     });
-                    callback();
+                    context.executeAction(actionFive, {}, callback);
                 };
                 actionFour.displayName = 'Four';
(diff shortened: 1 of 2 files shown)


packages/fluxible/tests/unit/lib/FluxibleContext.js
```diff
@@ -522,9 +540,17 @@ describe('FluxibleContext', function () {
...
                 };
                 actionThree.displayName = 'Three';
+                var actionFour = function (context, payload, callback) {
+                    actionCalls.push({
+                        context: context,
+                        payload: payload
+                    });
+                    callback();
+                };
+                actionFour.displayName = 'Four';
```
Vijar: @mridgway can you also try with action four calling another deeper action?

`One.Four.Six` case

Vijar:
:ship:

yahoocla:
CLA is valid!

packages/fluxible/tests/unit/lib/FluxibleContext.js
Vijar: Nevermind

PR #313, opened 2015-11-19
Update Components.md

dbillingham:
TestComponent's connectToStores state getter is trying to access FooStore. connectToStore no longer passes the stores to the state getter. The state getter signature is now (context, props) and you should access the store using `context.getStore(FooStore)`.


mridgway:
Thank you!

PR #315, opened 2015-11-19
Correct plugin example usage

rwoody:
The plugin example usage appears to have fallen out of date with the example `TestPlugin` code. Specifically, `getFoo` is provided as context to retrieve the `foo` value passed into `Fluxible.createContext`.

I've also added `contextTypes` to the Application element in an effort to help beginners with a more complete example.


mridgway:
:+1:

yahoocla:
CLA is valid!

Issue #316, opened 2015-11-25
[addons-react] Update FluxibleMixin to check for autobind:false


redonkulus:
If set to false, then manually bind the onChange handler.

https://github.com/yahoo/fluxible/blob/master/packages/fluxible-addons-react/FluxibleMixin.js#L116

Vijar:
@mridgway thoughts?

I feel like we can avoid having users do the current workaround of 
```js
componentWillMount: function () {
    this.onChange = this.onChange.bind(this);
}
```

mridgway:
Sure, I'm fine with this.

mridgway:
The `autobind` setting is not added to the React components, it is only used during construction of the React component class. Therefore it is currently not possible to detect whether autobinding is enabled or not.

ericf:
```
this.constructor.prototype.__reactAutoBindMap
```

ericf:
@mridgway and I discussed this, and came up with some options. I think the simplest might be to just set the context to `this` (the component instance)` when registering the event listener: https://github.com/yahoo/fluxible/blob/d26300e8add9fea3ac4ce25ea8c9526132872db1/packages/fluxible-addons-react/FluxibleMixin.js#L188

redonkulus:
Closing as mixins are being deprecated.

PR #317, opened 2015-11-30
Detect autobound components in FluxibleMixin

Vijar:
@mridgway @redonkulus


yahoocla:

yahoocla:
CLA is valid!

Vijar:
Actually turns out that we can't really read `this.autobind` as a static property so this would never work...

I'll probably just close this unless you have any more ideas @mridgway @redonkulus

mridgway:
Yup, that is never exposed on the class, only used as part of the spec in `createClass`. It also wouldn't be consistent since the main class may not be autobound but the mixins may be.

mridgway:
Since this will not work, I am closing this PR. We will have to think of something else if we want to resolve #316.

PR #320, opened 2015-12-08
Add event emitters to executeAction for debugging

mridgway:
This allows for debugging all execute actions that pass through the Fluxible context. Eventually we should add the same ability for `dispatch` calls, which would be part of dispatchr.

### DIFF
--- a/packages/fluxible/lib/FluxibleContext.js
+++ b/packages/fluxible/lib/FluxibleContext.js
@@ -7,6 +7,8 @@
 var debug = require('debug')('Fluxible:Context');
 var isPromise = require('is-promise');
 var generateUUID = require('../utils/generateUUID');
+var EventEmitter = require('eventemitter3');
+var inherits = require('inherits');
 
 var __DEV__ = process.env.NODE_ENV !== 'production';
 require('setimmediate');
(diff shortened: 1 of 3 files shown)


yahoocla:

yahoocla:
CLA is valid!

rkrauskopf:
Is there any update on when this will be pulled into the next Fluxible release/update? 

This would be really useful for me since I've been asked to provide metrics on the time between an action and a subsequent view render so that our team can identify performance bottlenecks in our product.

redonkulus:
@mridgway LGTM :+1: 

@Vijar want to check?

lingyan:
Should we put this behind debug flag?

Vijar:
@rkrauskopf I've been experimenting with this internally and basically this is 99% there but doesnt really allow actions started by components to be measured very accurately(all NavLink->navigateActions).

So meanwhile, I am okay with merging this temporarily without documenting it. You can use it for your internal debugging, however we might need to change the api to a more complete implementation in the future without guaranteeing backwards compatibility. I just haven't had time to focus on this recently :(

rkrauskopf:
No problem! I found some other ways to do this that should be close to what I need. I'll still definitely use this feature for debugging though in the future. I understand you all are busy so thank you for taking the time to address this.

PR #322, opened 2015-12-11
Renaming this.listeners; adding some defensive code

src-code:
Renaming `this.listeners` to a namespaced property to avoid collisions, and adding some defensive code to ensure the property is an array as expected before trying to iterate on it, just to be safe.

cc @mridgway @redonkulus @Vijar

### DIFF
--- a/packages/fluxible-addons-react/FluxibleMixin.js
+++ b/packages/fluxible-addons-react/FluxibleMixin.js
@@ -49,7 +49,7 @@ var FluxibleMixin = {
      * @method componentDidMount
      */
     componentDidMount: function componentDidMount() {
-        this.listeners = [];
+        this._fluxible_listeners = [];
         var self = this;
 
         // Register static listeners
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

ericf:
The code looks good. This will help protect against one of the down sides of mixins: a single instance object mutated by different parties.

The one concern I have is with `listeners` previously _looking_ like a public property and people accessing it directly instead of going through `getListeners()`. If anyone is doing that, they'd be broken by this change.

src-code:
@ericf Yeah, that's true about the potential for breakage.  Not sure how likely that would be, as I'm not sure who'd need to access these listeners outside of the mixin itself. (There doesn't seem to be any cases of it within other parts of Fluxible at least, it seems.) @redonkulus @mridgway, any thoughts?

mridgway:
I'm ok with renaming it since the property is undocumented. :+1:

ericf:
:+1:

redonkulus:
:+1:

redonkulus:
@mridgway want to do a release?

PR #324, opened 2015-12-15
Add ability to batch `change` events

mridgway:
Opening a PR for discussion.

**Problem**: Change events can happen several times during a navigation causing confusing flashes of content. This also causes many re-renders which can cause performance issues.

**Solution**: Provide a way to batch change events from stores into a single React update. Allow the developer to decide when a transaction should begin and end.

**Implementation**

I have tried two implementations of a way to proxy all change events and buffer them until you are ready to flush them. Both implementations rely on a function that is called for each change event and the ability to start and end transactions.

One implementation is a higher-order component that attaches a manager to the React `context` which is then used to register listeners instead of directly on stores. It relies on being at the top of the rendering tree so that it encompasses all components that listen to stores. Also, it will only work with components that use `FluxibleMixin` or `connectToStores`.

The second implementation uses a Fluxible plugin to override `getStore` to return a store that has overridden `addChangeListener` and `removeChangeListener` methods that will use the manager instead. This would work for any implementation of store listening, but does feel hackier since it is modifying the store's methods.

### DIFF
--- a/packages/fluxible-addons-react/batchedUpdatePlugin.js
+++ b/packages/fluxible-addons-react/batchedUpdatePlugin.js
@@ -3,24 +3,97 @@
  * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
 'use strict';
+
 var batchedUpdates = require('react-dom').unstable_batchedUpdates;
 
+function ChangeManager (context, strategy) {
+    var self = this;
+    self.context = context;
+    self.strategy = strategy;
+    self.batch = null;
+    self.depth = 0;
+    self.transactionInterface = {
+        isStarted: function isStarted() {
+            return !!self.depth;
+        },
+        start: self.startTransaction.bind(self),
+        end: self.endTransaction.bind(self)
+    };
+}
+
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

mridgway:
So we just implemented this internally with the plugin implementation with pretty good success. From this implementation we learned the following:

 * transaction management should probably happen within a store so that it can be based on events that are dispatched
 * the store that manages the transactions bypass the transaction so that a component could show transitions or loading spinners during the transaction

With this I think I will remove the HoC implementation and continue iterating on the plugin.

mridgway:
Updated the PR to stick with using a plugin for this. This fits in well with the existing batchUpdatePlugin, so I have added this functionality there.

webcarrot:
My simple es6 plugin - async & await

http://pastebin.com/iBvjJ0Tm

redonkulus:
@mridgway its been a while, do you still want to pull this in, make a new batch plugin or just close as reference?

PR #344, opened 2015-12-18
Convert createClass calls to React.Component classes where possible; Remove autobinding otherwise

mridgway:
This should improve performance during instantiation of the components significantly.


redonkulus:
:+1:

yahoocla:
CLA is valid!

PR #350, opened 2016-01-08
client-side only fluxible app example

malaman:
Hi, 
for some reasons, we use browser only application with fluxible and fluxible-router.
I thought it might be interesting to other people how to bootstrap non-isomorphic application with fluxible, so i created small example.

### DIFF
--- a/examples/fluxible-router/README.md
+++ b/examples/fluxible-router/README.md
@@ -4,7 +4,7 @@ This shows routing on the client and the server with [Fluxible](http://fluxible.
 
 ## Usage
 
-Make sure you run `npm install` from the root of flux-examples. Then change directories to routing (`cd routing`) and run:
+Make sure you run `npm install` from the root of flux-examples. Then change directory to fluxible-router (`cd fluxible-router`) and run:
 
 ```bash
 $ npm run dev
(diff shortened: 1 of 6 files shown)


redonkulus:
Thanks. You might want to take a look at the chat example, that supports query params which can defer the rendering to the client: https://github.com/yahoo/fluxible/blob/master/examples/chat/server.js#L39

The readme has more information: https://github.com/yahoo/fluxible/tree/master/examples/chat#usage

yahoocla:
CLA is valid!

mridgway:
I'd prefer not to have a duplicate of the application just to change a couple of files for client-only. As @redonkulus mentioned, the chat example already has the ability to be client-only rendered or server-only rendered in addition to being isomorphic. Maybe we could apply the same logic to the fluxible-router example as well?

I see that your example just uses a .html page as well, maybe we could just add that to the other folders and document how to use that instead.

malaman:
Ok. I will try to apply the changes/comments to the fluxible-router example .

malaman:
I've modified contents of fluxible-router example and remove duplicated application.

mridgway:
:+1: Looks good to me.

redonkulus:
:+1:

Issue #349, opened 2016-01-07
[fluxible-router] Safari popstate issue causing a handleHistory.js to execute the navigateAction on page load.


skippykawakami:
Safari fires popstate on page load. Ordinarily this shouldn't cause an issue in fluxible-router, but because of how we're doing some ATS things, it ends up executing the navigateAction. Basically in our prod environment, `url` and `currentUrl` won't match inside `_onHistoryChange`, which causes the navigateAction to execute. `handleHistory.js` should guard against this.

For reference on the Safari issue:
* http://stackoverflow.com/questions/28441025/ignore-safaris-initial-page-load-popstate
* http://stackoverflow.com/questions/15896434/window-onpopstate-on-page-load
* https://github.com/visionmedia/page.js/issues/213

lingyan:
@skippykawakami Thanks for reporting.  Let me see what's the best way to work around this.

skippykawakami:
@lingyan Here's an alternate solution that avoids having to set a timeout. I'm sure I don't know all the scary edge-cases fluxible-router has to handle, so this approach may be a bad idea, but it seems to prevent this issue when I use it in the project I'm working on.

https://github.com/skippykawakami/fluxible/commit/e553f8784751d344894d5745aa7d934a94141ce5

lingyan:
@skippykawakami  That's a good approach.  The only thing is to test out this use case: load page 1 -> click and navigate to page 2 -> browser refresh reload page 2.  On the last refresh, check this window.history.state.  It might still have `fluxible: true` in there.

The other approach I was gonna try is to extract the `url === currentUrl` into a customizable function so you can customize.  

Also could you elaborate a little on the "ATS thing" you mentioned?  Is your ATS remapping the url to a different path, therefore `url` is not matching `currentUrl`?

skippykawakami:
@lingyan Yeah, the ATS mapping we have for our bucketing and dogfooding appends a query param to the url it's requesting for some reason (I believe it's a well thought-out reason, but I'm not sure what it is). 

I'll check your use case and see what I find. If that really is the case, it might be better if I check for the fluxible=true in the popState event object's state property rather than the history property (actually maybe that's better overall... Surely there's nothing in e.state for a back button that no one actually pressed)

skippykawakami:
@lingyan You were right about reloading being an issue. Checking against the event object instead of window.history seems to fix it.  https://github.com/yahoo/fluxible/pull/351

lingyan:
[...]

PR #351, opened 2016-01-08
Safari fix

skippykawakami:
@lingyan Good call with that use case you mentioned (involving reloading the page after a pushState). I changed the flexible: true check to check the popstate event object instead.

Anyhow, here's a PR if this ends up being the approach you want to take.


yahoocla:

yahoocla:
CLA is valid!

lingyan:
@skippykawakami 

I don't have Safari 8 any more, so can't test with that version.  But I do see the same popstate fired on page load issue on Safari 9.

There is timing issue with this approach:
* Sometimes, `_onHistoryChange` is [attached](https://github.com/skippykawakami/fluxible/blob/safari-fix/packages/fluxible-router/lib/handleHistory.js#L104) after the popstate event is fired, which is what your change is trying to fix.
* Other times, it is attached before popstate event is fired.  In this case, `url` and `currentUrl` still won't match.

Let me think if there is a better solution for this, other than what I mentioned in the issue:
* change ATS to pass additional info via header, instead of query param: https://github.com/yahoo/fluxible/issues/349#issuecomment-170201940
* or, extract the `url === currentUrl` logic into a customizable function so apps can customize

skippykawakami:
@lingyan 

Hmm... yeah, I see the potential timing issue. I should point out that my change doesn't suppress the popstate event, it just let's us know when we can ignore it. However I think I can make a few changes here that may handle things when `_onHistoryChange` is attached before the false popstate event fires. Basically, we add the same `state.fluxible` check to `_onHistoryChange` that we have in the preload listener. And in `componentDidMount` where we check if there's a pending popstate event, we `replaceState` with the current url if there's no pending event. This lets us safely ignore popstate events that don't have `state.fluxible` without worrying that we're missing the original page request. This also properly populates `origUrl`

lingyan:
@skippykawakami check out my alternative solution in #353

skippykawakami:
@lingyan Hey, didn't see you'd created a PR until I updated mine. I'll check it out.

lingyan:
@skippykawakami haha, no problem. Let me know what you think. thanks!

PR #353, opened 2016-01-13
check currentNavigate.externalUrl for url matching to address #349

lingyan:
@skippykawakami

cc: @mridgway @redonkulus 

An alternative to #351 for fixing a Safari edge case reported in #349.  Instead of messing with detecting popstate event on page load for safari browser only (which requires user agent parsing that I want to avoid), or messing with browser history, this approach is safer.

##### Work needed on app side:
* dispatch `externalUrl` as part of the payload for the NAVIGATE_START event on server side.  For client side navigation, `externalUrl` field is not needed and will not be set.

##### Notes:
* I would still strongly suggest you to get your ATS team to change to pass HTTP header instead of query param, if they can :)
* I haven't updated doc for this, because I'd like to keep this as an undocumented feature for now.

Give it a try and let me know what you think.  Thanks!

### DIFF
--- a/packages/fluxible-router/lib/handleHistory.js
+++ b/packages/fluxible-router/lib/handleHistory.js
@@ -88,7 +88,7 @@ function createComponent(Component, opts) {
                 //   for the route represented by the hash fragment.
 
                 var urlFromHistory = this._history.getUrl();
-                var urlFromState = this.props.currentRoute && this.props.currentRoute.get('url');
+                var urlFromState = this.props.currentRoute && this.props.currentRoute.url;
 
                 if ((urlFromHistory !== urlFromState)) {
                     debug('pageload navigate to actual route', urlFromHistory, urlFromState);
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

packages/fluxible-router/lib/handleHistory.js
```diff
@@ -88,7 +88,7 @@ function createComponent(Component, opts) {
                 //   for the route represented by the hash fragment.
 
                 var urlFromHistory = this._history.getUrl();
-                var urlFromState = this.props.currentRoute && this.props.currentRoute.get('url');
+                var urlFromState = this.props.currentRoute && this.props.currentRoute.url;
```
lingyan: this is a bug fix not directly related to #349

packages/fluxible-router/lib/handleHistory.js
```diff
@@ -121,14 +121,21 @@ function createComponent(Component, opts) {
...
-            var currentRouteMap = props.currentRoute || {};
-            var currentUrl = currentRouteMap.url;
+            var currentRoute = props.currentRoute || {};
+            var nav = props.currentNavigate || {};
+
+            // Add currentNavigate.clientRelativeUrl for https://github.com/yahoo/fluxible/issues/349:
+            // "Safari popstate issue causing handleHistory.js to execute the navigateAction on page load".
+            // This needs app to dispatch "clientRelativeUrl" as part of the payload for the NAVIGATE_START event
+            // on server side.  For client side navigation, this field is not needed and will not be set.
+            var currentUrl = nav.clientRelativeUrl || currentRoute.url;
```
lingyan: @skippykawakami For example, if request url in browser is http://mydomain.com/abc, and ATS changes it to http://mydomain.com/abc?a=1 for whatever reason, (**NOTE** using query param is not recommended.  should use HTTP header instead.), your app can dispatch `NAVIGATE_START` event with payload like this:
```
{
   url: '/abc',
  method: 'GET',
  clientRelativeUrl: '/abc?a=1'
}
```

packages/fluxible-router/lib/handleHistory.js
mridgway: We should do this in a separate PR.

mridgway: Nvm, looks like it was already done.

kaesonho:
:+1:

Issue #355, opened 2016-01-14
Please clarify interface of fluxible store


c089:
I am a bit confused by the documentation of stores:

For a `Store` it is "recommended to extend an event emitter so that your store can emit change events to the components". This is also what the examples for writing your own store do, and how [BaseStore](https://github.com/yahoo/fluxible/blob/master/packages/dispatchr/addons/BaseStore.js) and [createReducerStore](https://github.com/yahoo/fluxible/blob/master/packages/fluxible-reducer-store/docs/api/createReducerStore.md) (indirectly by reusing [createStore](https://github.com/yahoo/fluxible/blob/master/packages/dispatchr/addons/createStore.js)) are implemented. But then `connectToStore` uses `addChangeListener` on the store, which in turn does nothing but delegate to `on('change')`. Yet that just seems to be copied from the [Flux docs](https://facebook.github.io/flux/docs/overview.html#structure-and-data-flow) which also say: 

> The stores then emit a _change_ event to alert the controller-views that a change to the data layer has occurred.

So what is fluxible store API now? `store.AddChangeListener` or `store.on('change'`)? At the moment both work, but I would like to make sure I'm using the API right.

mridgway:
`store.addChangeListener()` is a convenience method added by BaseStore and createStore. This follows the same API that Flux actually used: https://github.com/facebook/flux/blob/e0a52d0dada7b78e7197026dc8cf35d84991f0e8/examples/flux-chat/js/stores/UnreadThreadStore.js#L32

These utilities are not required to use Fluxible: you can implement your own stores without extending Fluxible's as basic classes. The [documentation on Stores](http://fluxible.io/api/stores.html) documents the required API for building your own classes and the [`BaseStore` docs](http://fluxible.io/addons/BaseStore.html#built-in-methods) talk about the methods that are added on top of the required store API.

I think we could clarify this in the docs better. If you have any ideas on how to do that please let me know.

mridgway:
> But then connectToStore uses addChangeListener on the store, which in turn does nothing but delegate to on('change').

We should probably change this to use `on('change', listener')` instead so that it works with stores that don't extend `BaseStore`. I will create a separate issue to track this.

mridgway:
Created https://github.com/yahoo/fluxible/issues/356 as a followup for connectToStores. 

This issue can be used for documentation changes that are needed.

c089:
> I think we could clarify this in the docs better. If you have any ideas on how to do that please let me know.

[...]

PR #357, opened 2016-01-16
Consistently use the public API for `Store`

### LINKED ISSUE #356: [fluxible-addons-react] connectToStores expects `addChangeListener` method
https://github.com/yahoo/fluxible/blob/master/packages/fluxible-addons-react/connectToStores.js#L34

This should use the `.on('change', listener)` instead since it is not required to have the `addChangeListener` method if store does not extend `BaseStore`.

c089:
As discussed in #355, `addChangeListener` / `remvoeChangeListener` are
convenience methods added by `BaseStore` and `createStore`, but not
necessarily defined on any instance of `Store`. To ensure
`connectToStores` and `FluxibleMixin` work with every store instance, we
use the `EventEmitter` API to listen for `change` events instead.

Fixes #356.

### DIFF
--- a/packages/fluxible-addons-react/FluxibleMixin.js
+++ b/packages/fluxible-addons-react/FluxibleMixin.js
@@ -185,7 +185,7 @@ var FluxibleMixin = {
             throw new Error('storeListener mixin called listen when component wasn\'t mounted.');
         }
 
-        listener.store.addChangeListener(listener.handler);
+        listener.store.on('change', listener.handler);
         this._fluxible_listeners.push(listener);
     },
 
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

redonkulus:
:+1:

mridgway:
Thanks for the PR.

One thing to add: the `removeChangeListener(listener)` calls should also be updated to `removeListener('change', listener')` since `removeChangeListener` is also added by `BaseStore`.

c089:
Oh yes of course, thanks for that. I amended the commit :)

mridgway:
:+1:  Thanks @c089!

PR #360, opened 2016-01-20
Allow navParams to be overwritten by overwriteSpec

snyamathi:
@lingyan 

This exposes getNavParams to the overwriteSpec so that it can be overwritten should a component need to set custom navParams during one of the other event hooks.


yahoocla:

yahoocla:
CLA is valid!

lingyan:
:+1:

kaesonho:
same as @lingyan , then :+1:

lingyan:
:+1:

PR #362, opened 2016-01-25
Add getStore to storeContext

webcarrot:
Plugins working with plugStoreContext have no access to getStore function.
I added it.

PS. ugly and tested this time...


yahoocla:

yahoocla:
CLA is valid!

mridgway:
One way that you could get access to this is through the second parameter on the `plugContext` method.

```js
//...
plugContext: function (options, context) {
    return {
        plugStoreContext: function (storeContext) {
            context.getStore(...);
            //...
        }
    };
}
//...
```

webcarrot:
Ok.

PR #364, opened 2016-01-27
update react-router example to use react-router 1.0

hfter789:


yahoocla:

yahoocla:
CLA is valid!

mridgway:
Nice. Thanks for getting this started.

examples/react-router/server.js
```diff
@@ -12,39 +12,49 @@ var React = require('react');
...
 
 server.use(function (req, res, next) {
-    var context = app.createContext();
 
-    debug('Executing navigate action');
-    Router.run(app.getComponent(), req.path, function (Handler, state) {
-        context.executeAction(navigateAction, state, function () {
-            debug('Exposing context state');
-            var exposed = 'window.App=' + serialize(app.dehydrate(context)) + ';';
+	debug('Executing navigate action');
```
mridgway: Could you switch to using spaces instead of tabs in this file?

examples/react-router/package.json
```diff
@@ -12,7 +12,8 @@
   "dependencies": {
     "fluxible": "^0.5.0",
     "fluxible-addons-react": "^0.1.6",
+    "history": "^1.17.0",
     "react": "^0.13.0",
-    "react-router": "^0.13.1"
+    "react-router": "^1.0.0"
```
mridgway: We should move these dependencies to the [parent directory's package.json](https://github.com/yahoo/fluxible/blob/master/examples/package.json#L22). This way there isn't an additional install needed for this example anymore.

hfter789: Good point. 
After I moved the dependencies under examples/package.json, react-router server cannot correctly match the url, and there is no error. I will investigate more later.

examples/package.json
```diff
@@ -17,6 +17,7 @@
     "fluxible-addons-react": "^0.2.0",
     "fluxible-plugin-fetchr": "^0.3.3",
     "fluxible-router": "^0.4.0",
+    "history": "^1.17.0",
     "react": "0.14.x",
     "react-dom": "0.14.x",
     "react-router": "^1.0.0-rc3",
```
mridgway: Update to `^1.0.3` here. I'm seeing weird issues with this where it's installing `1.0.2` and throwing peer invalid errors on installation.

hfter789:
@mridgway Changed the version number, but the problem still exists.

hfter789:
@mridgway This PR is ready for review. And thank you for all the help.

mridgway:
:+1:

redonkulus:
:+1:

mridgway:
Forgot to say, thanks a bunch @hfter789!

PR #368, opened 2016-02-03
not check matched route before updating currentRoute

kaesonho:
@mridgway @lingyan @redonkulus 

similar as #367 

the PR fix an issue that we have `navigate` object, which includes `transection id`, it's not updated when we navigate to the same url. 

@mridgway instead of having a `else` condition to just update `navigate` id, the proposal of this PR is just updating currentRoute anyway. otherwise when we navigate to the same url, the flow still finishes (i.e., we have NAVIGATE_START, NAVIGATE_SUCCESS ...everything), but `currentRoute` is not updated , might cause some issue if url is keeping the same but with different `params/query`.

### DIFF
--- a/packages/fluxible-router/lib/RouteStore.js
+++ b/packages/fluxible-router/lib/RouteStore.js
@@ -33,10 +33,7 @@ var RouteStore = createStore({
             method: navigate.method
         });
 
-        if (!this._areEqual(matchedRoute, this._currentRoute)) {
-            this._currentRoute = matchedRoute;
-        }
-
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

lingyan:
:+1:

redonkulus:
:+1:

gfranko:
:+1: :100:

mridgway:
Published in fluxible-router@0.4.7

Issue #337, opened 2015-12-17
[router] NavLink should detect whether to follow link automatically


mridgway:
See https://github.com/yahoo/fluxible-router/issues/44

hfter789:
@mridgway Is this done? I am seeing that the followLink prop https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/createNavLinkComponent.js#L109 is doing the job.

mridgway:
This feature would automatically check if the route is registered and follow the link if it is not.

mridgway:
Added in #396

PR #369, opened 2016-02-05
Follow NavLinks that do not match a registered route

hfter789:
Attempt to solve this issue #337 
New logic on href field of NavLink:
- href is treated as routeName. if such routeName doesn't exist in route, use href directly.

### DIFF
--- a/examples/fluxible-router/components/Nav.js
+++ b/examples/fluxible-router/components/Nav.js
@@ -10,6 +10,9 @@ function Nav () {
         <ul className="pure-menu pure-menu-open pure-menu-horizontal">
             <li><NavLink routeName="home" activeStyle={{backgroundColor: '#ccc'}}>Home</NavLink></li>
             <li><NavLink routeName="about" activeStyle={{backgroundColor: '#ccc'}}>About</NavLink></li>
+            <li><NavLink href="/page/Relative" activeStyle={{backgroundColor: '#ccc'}}>Relative Href</NavLink></li>
+            <li><NavLink autoMatch={true} href="internal" activeStyle={{backgroundColor: '#ccc'}}>AutoMatch Href Internal</NavLink></li>
+            <li><NavLink autoMatch={true} href="//www.yahoo.com" activeStyle={{backgroundColor: '#ccc'}}>AutoMatch Href External</NavLink></li>
         </ul>
     );
 }
--- a/examples/fluxible-router/configs/routes.js
+++ b/examples/fluxible-router/configs/routes.js
@@ -23,6 +23,17 @@ export default {
             done();
         }
     },
+    internal: {
+        path: '/internalPage',
+        method: 'get',
+        handler: require('../components/Page'),
+        label: 'Internal',
+        action: (context, payload, done) => {
+            context.dispatch('LOAD_PAGE', { id: 'Internal Page' });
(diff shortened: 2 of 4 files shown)


yahoocla:

yahoocla:
CLA is valid!

mridgway:
I don't think this is exactly what we had in mind. #337 is about making the router know that a given `href` is a registered route or not and determining whether to do a pushState navigation or let the browser handle the navigation.

hfter789:
Hi @mridgway. Thanks for the feedback. I understand your logic. In this case, I can remove the flag check to have routeName as a fallback https://github.com/hfter789/fluxible/blob/a9d97de3573b64a8c1e542a2d251d2d703a88cd8/packages/fluxible-router/lib/createNavLinkComponent.js#L92. would look like this:
`href = routeStore.makePath(routeName, navParams) ||  routeName;`
Since makePath checks whether routeName is registered in the router (https://github.com/yahoo/routr/blob/5d4ed4b65f95f9737621380e0d25cab4cc1a1d89/lib/router.js#L260), I know that routeName is external if makePath returns null.

 I use the followLink not because I want to tell NavLink this is an external link, but instead, I want to tell NavLink that this **could** be an external link. With this flag, users that don't expect to have external link would not navigate to external link accidentally because of typos. And users that don't know if they would have external links can add this flag to all their NavLink, and then NavLink can fallback when route is not registered.

Please let me know if I got anything wrong. I can modify the PR to make it work.

mridgway:
I see what you're saying. I guess I just didn't understand from the example that you added. Could you add a test case for this so that we can ensure we don't break this?

mridgway:
I'm not sure if `routeName` is what we should be using for this though. I still think that `href` is where you would put the external link and then we'd detect if it maps to a registered route pattern. We already check the href to see if it has a separate domain, but not whether for instance, `/login` matches a registered pattern.

hfter789:
I see. It seems like routeName should be expected to register to a router. On the other hand, href could possibly be registered to a router or external. Am I right?

mridgway:
@hfter789 Yes, exactly.

hfter789:
Hi @mridgway. I updated my PR. Can you give me some feedback on it? I can add some unit tests if the logic looks good to you.

mridgway:
Looks good to me @hfter789. That's exactly the functionality I was thinking of. :+1:  on the tests.

mridgway:
We also may need to find some way to make this configurable so that we don't break existing functionality in applications that have client-side 404 handling.

hfter789:
@mridgway You are right. I would add a config to NavLink and unit tests.

hfter789:
Hi @mridgway I updated my PR, please take a look.

hfter789:
@mridgway Fixed.

mridgway:
Sorry for the delay. Circling back around to this. Looking good. One question: the support for converting `href` from a route name to a path doesn't make sense to me. What is the use case for this and why wouldn't you just use `routeName` instead of `href`?

hfter789:
Hi @mridgway I am just grabbing issues from the issue board. I think the issue might explain the context better.

mridgway:
Ok, I'm going to make a couple of changes, add documentation, and merge this.

hfter789:
:+1: Thanks.  Let me know if you need anything.

mridgway:
Moved to #396.

PR #375, opened 2016-02-17
Fix overflow scroll bar issue

tom76kimo:
Current search bar:
[https://gyazo.com/c28ff3f3527c7ae0fc4d8ed1bc5540f1](https://gyazo.com/c28ff3f3527c7ae0fc4d8ed1bc5540f1)

I think `overflow:hidden` would be enough.

### DIFF
--- a/site/components/Search.jsx
+++ b/site/components/Search.jsx
@@ -88,7 +88,7 @@ class Search extends React.Component {
         let classes = cx({
             'D(ib) Mstart(3px) Ov(h) Va(m) Pos(a) End(20px)': true,
             'W(0)': this.state.visible === false,
-            'W(200px) Ov(h)': this.state.visible
+            'W(200px)': this.state.visible
         });
         return (
             <div className="D(ib)">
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

redonkulus:
@tom76kimo thanks for the PR. Which browser and platform are you seeing this on?

tom76kimo:
I use `Chrome 48.0.2564.109` and `OSX 10.9.5`

site/components/Search.jsx
```diff
@@ -88,7 +88,7 @@ class Search extends React.Component {
         let classes = cx({
             'D(ib) Mstart(3px) Ov(h) Va(m) Pos(a) End(20px)': true,
             'W(0)': this.state.visible === false,
-            'W(200px) Ov(a)': this.state.visible
+            'W(200px) Ov(h)': this.state.visible
```
mridgway: Looks like we can just remove the `Ov(h)` here since its already specified in the `true` case.

mridgway:
I haven't been able to reproduce, but I don't see the harm in this. I added a comment above, looks like it's duplicating the `Ov(h)`, so if you want to remove that we'll merge this in.

tom76kimo:
You're right! I removed the duplicated class.

mridgway:
Thanks!

tom76kimo:
:+1:

Issue #377, opened 2016-02-19
do not assign ref to stateless function components


pwmckenna:
On React 0.14 I get the following warning when using `connectToStores` on a stateless function component:
 
```
Warning: Stateless function components cannot be given refs (See ref "wrappedElement" in StatelessComponent created by StoreConnector). Attempts to access this ref will fail.
```
when doing something like

```js
connectToStores(() => <div>Hello World</div>, [Store], context => ({}));
```

because you can't specify a ref for a stateless component.See the note at the bottom of  https://facebook.github.io/react/docs/reusable-components.html#stateless-functions

# Suggested Fix:

Use a check like this:
https://github.com/acdlite/recompose/blob/master/src/packages/recompose/isClassComponent.js

to conditionally pass in the ref here:
https://github.com/yahoo/fluxible/blob/master/packages/fluxible-addons-react/connectToStores.js#L56

mridgway:
https://github.com/facebook/react/issues/4936 for more context.

If this check is reliable, :+1: on adding it to all HoCs.

pwmckenna:
I'm happy to put together but I immediately was stumped by the choice of where to get the `isComponentClass` function.

- shared utility in fluxible...not sure how to do this given that multiple repos need it (fluxible, fluxible-router, etc)
- require recompose
- pull isComponentClass into its own repo, publish, then have each repo require it.
- inline the function where needed.

Take your pick and I'll implement :)

mridgway:
I looked into whether React has its own function for checking this, but it seems they just inline it: https://github.com/facebook/react/blob/bbd5a78efaf92b7ffcde2a1923112b8931855209/src/renderers/shared/reconciler/ReactCompositeComponent.js#L164 for the best performance. I think we can just use the same logic since HoC's always require a `Component` to be passed in anyway.

pwmckenna:
sounds good, i'll make a PR

PR #378, opened 2016-02-22
do not assign pure function components a ref

pwmckenna:
happy to adjust line length stuff, or c+p those tests into each package that does this...or anything else really. :evergreen_tree: 

fixes https://github.com/yahoo/fluxible/issues/377

### DIFF
--- a/packages/fluxible-addons-react/connectToStores.js
+++ b/packages/fluxible-addons-react/connectToStores.js
@@ -53,7 +53,8 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
             }
         },
         render: function render() {
-            return React.createElement(Component, Object.assign({}, this.props, this.state, {ref: 'wrappedElement'}));
+            var props = Component.prototype && Component.prototype.isReactComponent ? {ref: 'wrappedElement'} : null;
+            return React.createElement(Component, Object.assign({}, this.props, this.state, props));
         }
     });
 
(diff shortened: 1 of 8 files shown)


yahoocla:

yahoocla:
CLA is valid!

packages/fluxible-addons-react/connectToStores.js
```diff
@@ -53,7 +53,7 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
             }
         },
         render: function render() {
-            return React.createElement(Component, Object.assign({}, this.props, this.state, {ref: 'wrappedElement'}));
+            return React.createElement(Component, Object.assign({}, this.props, this.state, Component.prototype && Component.prototype.isReactComponent ? {ref: 'wrappedElement'} : null));
```
mridgway: Let's pull this out into a separate variable:

`var props = Component.prototype && Component.prototype.isReactComponent ? {ref: 'wrappedElement'} : null;`

mridgway:
Looks good. We should pull the props out to a separate variable. And yeah, we should have a test for each component if you don't mind.

packages/fluxible-addons-react/connectToStores.js
gingur: +1 for breaking logic into separate variable.

pwmckenna:
updated with seperate `props` variables, and added tests for each usage. There was some variation between how the packages tested, so I tried to conform to each of the packages' styles.

mridgway:
:+1: awesome work @pwmckenna! Will wait for tests to pass and get this merged.

PR #385, opened 2016-03-05
Update dev dependencies

### LINKED ISSUE #300: [generator] Upgrade to Babel 6
Any plan to upgrade to babel 6 ?

mridgway:
Resolves #300

This upgrades the development environment for Fluxible to Node 4+, Babel 6, and eslint 2 as well as updating other dev dependencies.

It's important to note that Babel 6 does not support decorators anymore (I tried the `babel-plugin-transform-decorators-legacy` but it does not work the same way as the old decorator implementation) so I have removed the tests for this functionality (although it still exists for babel 5 users).


redonkulus:
:+1:

yahoocla:
CLA is valid!

PR #392, opened 2016-03-09
stopPropagation in all cases

Diokuz:
Move stopPropagation to top of the handler, to prevent propagation in case of modified event or right click event.

Steps to repro:

1. Place NavLink inside an element, which have onClick handler.
2. Set `stopPropagation` attribute of NavLink to `true`.
3. Press CMD+LM or RM.

Expected: propagation stopped, outer element onClick handler never called.

Actual: propagation did not stopped, onClick on outer element fired.


yahoocla:

yahoocla:
CLA is valid!

mridgway:
Thanks!

PR #394, opened 2016-03-11
allow ignore popstate event on page load, this is mainly a fix for sa…

lingyan:
…fari browser, see discussions in issue #349 and PR #351 for more details

@mridgway @redonkulus @kaesonho 
cc: @skippykawakami 

This PR does not change existing behavior, by introducing the fix behind a feature flag `ignorePopstateOnPageLoad`, with default value `false`.  Most of the apps will not need to enable this.  It is only needed for a special case where the url in currentRoute (in dehydrated RouteStore) is different from what is in the browser, typical due to an intermediate layer like ATS altering url (adding query parameter whatnot) before sending request to origin server.

The mechanism for identifying pageload popstate event:
* In Safari browser, when page loads, browser will trigger a popstate event with `null` state (in `event.state`)
* But we can't just check for `event.state === null` of popstate events to say it is triggered by page load, because if you load a new page in a browser, navigate to a new page, and on back button, a popstate event will be triggered, this event's `state` property will be `null`.
* To get around that, we manually replace the history state on page load (in componentDidMount() to be exact).  This way, all pages loaded with fluxible-router or client-navigated with fluxible-router will guarantee to have a non-null history.state object.  So all regular popstate events will guarantee to have a non-null `event.state`.

With that, we can use the following logic to tell whether the popstate event is triggered on page load:
```
                var stateFromHistory = this._history.getState();
                var isPageloadPopstate = (e.state === null) && !!stateFromHistory;
                debug('history listener detecting pageload popstate', e.state, stateFromHistory);
                if (isPageloadPopstate) {
                    debug('history listener skipped pageload popstate');
                    return;
                }
```

Some other alternatives for this issue is to delay `window.addEventListener('popstate', ...)` to after onload event.  Did not want to go that route, since that could introduce a delay comparing to what is currently supported.

### DIFF
--- a/packages/fluxible-router/lib/handleHistory.js
+++ b/packages/fluxible-router/lib/handleHistory.js
@@ -5,7 +5,7 @@
 /*global window */
 'use strict';
 var React = require('react');
-var debug = require('debug')('RoutingContainer');
+var debug = require('debug')('FluxibleRouter:handleHistory');
 var handleRoute = require('../lib/handleRoute');
 var navigateAction = require('../lib/navigateAction');
 var History = require('./History');
@@ -22,7 +22,8 @@ var defaultOptions = {
     enableScroll: true,
     historyCreator: function () {
         return new History();
-    }
+    },
+    ignorePopstateOnPageLoad: false
 };
 
 // Begin listening for popstate so they are not missed prior to instantiation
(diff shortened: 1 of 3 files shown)


packages/fluxible-router/lib/handleHistory.js
```diff
@@ -119,22 +137,30 @@ function createComponent(Component, opts) {
...
-            // "Safari popstate issue causing handleHistory.js to execute the navigateAction on page load".
-            // This needs app to dispatch "externalUrl" as part of the payload for the NAVIGATE_START event
-            // on server side, which contains the absolute url user sees in browser when the request is made.
-            // For client side navigation, "externalUrl" field is not needed and is not set by fluxible-router.
-            var externalUrl = nav.externalUrl;
-            if (externalUrl && externalUrl === window.location.href.split('#')[0]) {
-                // this is the initial page load, omit the popstate event erroneously fired by Safari browsers.
-                return;
-            }
-
```
lingyan: the check for `externalUrl` is no longer needed

mridgway:
:+1:

redonkulus:
:+1:

kaesonho:
:+1:

PR #398, opened 2016-03-18
Remove default active props and only listen to RouteStore when one is passed in

mridgway:
This is a breaking change.

This removes the default `activeClass` property value. This allows us to optimize NavLinks that do not use this feature: listening will only occur if one of the active properties (`activeClass`, `activeElement`, or `activeStyle`) is used.


yahoocla:

yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #399, opened 2016-03-20
use the container display name convention used by recompose

pwmckenna:
https://github.com/acdlite/recompose/blob/master/docs/API.md#wrapdisplayname

I don't really prefer one convention over another, but combining both has resulted in some pretty weird displayNames. this would just make my debugging life a tad easier :)

### DIFF
--- a/packages/fluxible-addons-react/connectToStores.js
+++ b/packages/fluxible-addons-react/connectToStores.js
@@ -23,7 +23,7 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
 
     inherits(StoreConnector, React.Component);
 
-    StoreConnector.displayName = componentName + 'StoreConnector';
+    StoreConnector.displayName = 'storeConnector(' + componentName + ')';
     StoreConnector.contextTypes = componentContextTypes;
 
     Object.assign(StoreConnector.prototype, {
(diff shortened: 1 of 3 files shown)


yahoocla:

yahoocla:
CLA is valid!

pwmckenna:
also, in a previous pr, i hard coded the implementation of [isClassComponent](https://github.com/acdlite/recompose/blob/master/docs/API.md#isclasscomponent). not a big deal, but if a dependency on recompose was acceptable we could probably abstract away a lot of the not-pure-fluxible stuff.

mridgway:
I'm fine with this. There are two tests in fluxible-addons-react `provideContext` that will need to be updated.

I think we'll hold off on the `isClassComponent` for now, but I see your point on the trend ;)

pwmckenna:
updated. thanks for the hint about the tests.

mridgway:
Thanks @pwmckenna! I've been meaning to do this for a while.

mridgway:
This is now published in v0.2.7.

pwmckenna:
@mridgway super sorry, broke contextProvider in 0.2.7. Here's the fix: https://github.com/yahoo/fluxible/pull/402/files

PR #400, opened 2016-03-22
[router] Inverts the navigate completion event payload

mridgway:
This structures the RouteStore state such that the navigate object contains the relevant metadata: route, error, isComplete. This allows us to get closer to an immutable store, but does change the object that is received by NAVIGATE_SUCCESS and NAVIGATE_FAILURE events:

 * NAVIGATE_SUCCESS received the route objet which contained `navigate`, but that is no longer the case. In order to relate the SUCCESS to the START event, it is required to pass in a `navigate` object instead.

 * NAVIGATE_FAILURE received an error object, but to maintain the similar interface to SUCCESS, this has been updated to the navigate object that contains addtional error information.

### DIFF
--- a/packages/fluxible-router/lib/RouteStore.js
+++ b/packages/fluxible-router/lib/RouteStore.js
@@ -17,53 +17,64 @@ var RouteStore = createStore({
     },
     initialize: function () {
         this._routes = null;
-        this._currentUrl = null;
-        this._currentRoute = null;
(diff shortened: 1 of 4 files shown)


yahoocla:

yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #401, opened 2016-03-22
Only store the currentRoute and use the immutable route for checking …

mridgway:
…shouldComponentUpdate

`href` and `isActive` are no longer computed on every prop/state change. Instead they are only computed if the props or `currentRoute` have changed.


redonkulus:
:+1:

roderickhsiao:
:+1:

PR #407, opened 2016-03-28
[resolves #405] Remove `payload` parameter in BaseStore.emitChange

### LINKED ISSUE #405: [dispatchr] Proposal: remove support for `emitChange` payloads in BaseStore
Payloads being emitted from `emitChange` can cause issues where state is actually being emitted rather than letting the store be used as the source of truth. This is problematic because a component may not be set up to listen to a store when it `emit`s the event, then after it has missed the event it renders based on the state of the store.

In short, this prevents imperative view changes based on events and forces developers to use the store state instead.

This would be a major breaking change for people that use these payloads. I'd be curious to hear use cases where they are being used for something that could not be done using store state.

mridgway:
See #405.

### DIFF
--- a/packages/dispatchr/addons/BaseStore.js
+++ b/packages/dispatchr/addons/BaseStore.js
@@ -65,11 +65,10 @@ BaseStore.prototype.shouldDehydrate = function shouldDehydrate() {
 /**
  * Emit a change event
  * @method emitChange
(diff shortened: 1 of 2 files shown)


redonkulus:
:+1:

yahoocla:
CLA is valid!

PR #410, opened 2016-03-31
Add query in some missing places

irae:
I was trying to do some advanced routing and I found out a few places it looks like the query might be useful.

Turns out I managed to do what I need without this changes. But I do think this might be a good idea for fluxible-router 1.0.0 because now you support query string on the NavLink. Suppose a user uses NavLink with the same route params, the history might be badly handled.

Feel free to close the Pull Request if this is not useful or if it is a mistake. I didn't fully test but this was the easier way to send a diff of the changes I wanted to suggest.

### DIFF
--- a/packages/fluxible-router/lib/navigateAction.js
+++ b/packages/fluxible-router/lib/navigateAction.js
@@ -12,7 +12,7 @@ function navigateAction (context, payload, done) {
         transactionId: context.rootId
     }, payload);
     if (!payload.url && payload.routeName) {
-        navigate.url = routeStore.makePath(payload.routeName, payload.params);
+        navigate.url = routeStore.makePath(payload.routeName, payload.params, payload.query);
         navigate.routeName = null;
     }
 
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

packages/fluxible-router/lib/navigateAction.js
```diff
@@ -12,7 +12,7 @@ function navigateAction (context, payload, done) {
         transactionId: context.rootId
     }, payload);
     if (!payload.url && payload.routeName) {
-        navigate.url = routeStore.makePath(payload.routeName, payload.params);
+        navigate.url = routeStore.makePath(payload.routeName, payload.params, payload.query);
```
irae: This is actually more important than the other file changes. makePath now supports query and only makes sense to accept query params on navigateAction payload.

mridgway: :+1: yup. This is a bug.

mridgway:
Creating a new PR with tests added.

PR #412, opened 2016-03-31
[resolves #411] Fix style attribute when NavLink is not active

### LINKED ISSUE #411: NavLink component no longer supports style property
I am very happy about query param support in NavLink, although style property no longer works.

https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/createNavLinkComponent.js#L207

I'd guess this line should be `var style = this.props.style` like it does for className.

mridgway:


yahoocla:

yahoocla:
CLA is valid!

lingyan:
:+1:

Vijar:
:+1:

redonkulus:
:+1:

PR #415, opened 2016-04-01
[resolves #414] Send correct payload to NAVIGATE_FAILURE

### LINKED ISSUE #414: NAVIGATE_FAILURE payload is still error object
Hi there,

In 1.0.0-alpha1, error from navigateAction doesn't complete navigation and isNavigateComplete property never becomes true.

I noticed that NAVIGATE_FAILURE receives error object but RouteStore expect it to be navigate object.
https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/navigateAction.js#L64

Cheers,

mridgway:
See #414


yahoocla:

yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #416, opened 2016-04-01
Action History

Vijar:
This just provides a method on context(`context.getActionHistory`) which returns structured debugging data for the action hierarchy. Everything is behind a context feature flag as well.

I'll provide incremental updates to this feature, including an open source action visualizer.

@redonkulus @mridgway @lingyan @kaesonho


yahoocla:

yahoocla:
CLA is valid!

packages/fluxible/lib/FluxibleContext.js
```diff
@@ -16,9 +16,10 @@ require('setimmediate');
  * A request or browser-session context
  * @class FluxibleContext
  * @param {Fluxible} app The Fluxible instance used to create the context
+ * @param {Object} [options]
```
redonkulus: please define what each option is

packages/fluxible/lib/FluxibleContext.js
```diff
@@ -152,7 +157,7 @@ FluxContext.prototype._initializeDispatcher = function initializeDispatcher() {
  * @param {Function} action The action to be executed to get the name from
  * @returns {Object}
  */
-FluxContext.prototype._createSubActionContext = function createSubActionContext(parentActionContext, action) {
+FluxContext.prototype._createSubActionContext = function createSubActionContext(parentActionContext, action, payload) {
```
redonkulus: doc block this

redonkulus:
I like it! We should add docs to the FluxibleContext API docs so people know its there and how to use it. We can possibly add this to the site docs as well.

Vijar:
@redonkulus @mridgway updated.

packages/fluxible/utils/callAction.js
```diff
@@ -13,7 +13,8 @@ require('setimmediate');
  * If done callback supplied, that indicates non-Promise invocation expectation,
  * otherwise, Promise invocation.
  */
-function callAction (actionContext, action, payload, done) {
+function callAction (context, actionContext, action, payload, done) {
```
lingyan: @Vijar Just curious, is `context` added for accessing `_enableDebug` only, or are you planning to make more use of it later?

Vijar: @lingyan just for the debug flag...do you think I should just pass in the flag?

lingyan: It's an internal function. Either way is fine with me.

packages/fluxible/utils/callAction.js
```diff
@@ -13,7 +13,8 @@ require('setimmediate');
...
+    var displayName = action.displayName || action.name;
```
mridgway: Is this used somewhere?

mridgway:
Ok, so the big question: is it possible for this to be a plugin instead of in core? I think you mentioned that measuring the `executeAction` calls from `componentContext` is not possible right now since it does not have a callback. Is that the only thing that you think would block that?

Vijar:
Closing in favor of #418

PR #417, opened 2016-04-04
Updates to routr 2.0.0

irae:
This expects https://github.com/yahoo/routr/pull/33 to be merged before merging this one.

I tested and everything works correctly and tests pass when linking the correct versions. Let me know if you want redundant tests to be added here for this particular use case or if keeping in the routr is good enough. IMHO it should be fine to just bump versions.


yahoocla:

yahoocla:
CLA is valid!

PR #421, opened 2016-04-08
[router] Fix scroll state in IE

mridgway:
@kfay


yahoocla:

yahoocla:
CLA is valid!

lingyan:
👍

kfay:
Thanks! :+1:

irae:
Just out of curiosity: Does fluxible router account for running inside of an `<iframe>`? One way to make it bullet proof is to use `element.ownerDocument.documentElement.offsetHeight` where element can be the element passed to `handleHistory`. `ownerDocument` in itself is IE6+ compatible, but were to pull the width/height from might be different across old browsers.

mridgway:
@irae No we do not account for iframes currently.

PR #423, opened 2016-04-11
provide overwriting getDefaultChildProps

kaesonho:
@mridgway @redonkulus @Vijar 

this adds the ability to dynamically decide the additional child props without wrap NavLink with another component.

also add JSDoc for all these functions with the same usage.

PR #424, opened 2016-04-12
provide getDefaultChildProps to fluxible-router-0.4.x

kaesonho:
@mridgway @redonkulus same as #423

### DIFF
--- a/packages/fluxible-router/lib/createNavLinkComponent.js
+++ b/packages/fluxible-router/lib/createNavLinkComponent.js
@@ -99,6 +99,9 @@ module.exports = function createNavLinkComponent (overwriteSpec) {
             }
             return href;
         },
+        getDefaultChildProps: function () {
+            return {};
+        },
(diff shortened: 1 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

mridgway:
:+1:

redonkulus:
@kaesonho can you document this feature somewhere?

mridgway:
Looks like the test you added failed.

kaesonho:
@redonkulus  right now we just say that we can overwrite things here  https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/docs/api/createNavLinkComponent.md,   

will add some more detail about what function we can overwrite in master   



BTW, should we change to use `ES6 extend` for this overwrite pattern?

kaesonho:
it's passing locally, re-run seems working, let me try to run couple times

PR #427, opened 2016-04-15
Catch pushState and replaceState exceptions

webcarrot:
Webkit on IOS throws exception: "SecurityError: DOM Exception 18" after many (~50) history changes.

Catch this error and failback to old method of url changing.

### DIFF
--- a/packages/fluxible-router/lib/History.js
+++ b/packages/fluxible-router/lib/History.js
@@ -103,7 +103,11 @@ History.prototype = {
 
             // remember the original url in state, so that it can be used by getUrl()
             var _state = Object.assign({origUrl: url}, state);
-            win.history.pushState(_state, title, url);
+            try {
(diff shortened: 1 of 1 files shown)


mridgway:
Is there a relevant webkit bug for this? I haven't experienced this.

yahoocla:
CLA is valid!

mridgway:
@ericf pointed me to https://bugs.webkit.org/show_bug.cgi?id=155901 and a https://trac.webkit.org/changeset/198687/ as a fix.

With that said. I'm fine with this fallback since it seems like a critical issue.

:+1:

ericf:
:+1: @mridgway when you merge this in do you want to add code comments with the links to the Webkit bugs so we don't forget next time we look at this code?

lingyan:
👍  thanks @webcarrot

PR #432, opened 2016-04-21
Ability to inspect dispatch calls as well as action/dispatch payloads

Vijar:
Tasks:
- [x] Add a `{...}` button which will output the payload to console
- [x] Allow ability to trace dispatch calls via `showDispatchCalls` prop
- [x] Hover on node, will show the link to it's parent action
- [x] Hover on `{...}` should show tooltip telling users to open console and click the `{...}` button


Vijar:
New UI:
screen shot 2016-04-21 at 4 24 20 pm

The dispatch calls are purple. I'm not sure how we want to show the color key though.

yahoocla:
CLA is valid!

Vijar:
@mridgway @redonkulus @ericf feedback appreciated

redonkulus:
👍

PR #433, opened 2016-04-22
connect the right plugin in the code!

ali1k:


redonkulus:
Good catch, thanks!

yahoocla:
CLA is valid!

PR #437, opened 2016-05-02
Point to #flux channel on Reactiflux

vcarl:
Discord has bad channel management, and as the #fluxible doesn't see much traffic, we're trying to point users where they're more likely to get some kind of answer. Also make it an instant join link so they're directed directly to the room.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This repository is the home of [Fluxible](http://fluxible.io/) and related libraries.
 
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

mridgway:
I think we're going to point our Readme to our gitter channel instead. The move to Discord seems to have killed off the engagement from our community.

PR #438, opened 2016-05-06
fix target:_blank vulnerability

yurited:
@lingyan @mridgway 
security fix for https://mathiasbynens.github.io/rel-noopener/


yahoocla:

yahoocla:
CLA is valid!

mridgway:
What if I wanted to use `target=_blank` and disable `noreferrer` and `noopener`? We would need to make an API to allow this if we want to make this the default. It may be confusing for users wanting that behavior to add new props for those flags.

For now I'm hesitant to change the default behavior of browsers since it makes the API unclean.

PR #442, opened 2016-05-12
added slidewiki use case

ali1k:


yahoocla:
CLA is valid!

mridgway:
Cool! 👍

PR #446, opened 2016-06-10
add a hint to getHandler error

samlecuyer:
`FluxibleMixin.js:120 Uncaught Error: storeListener attempted to add undefined handler. Make sure handlers actually exist.` is an extremely opaque error message and doesn't actually tell me how to fix it.  Adding a hint is useful.

@mridgway or @redonkulus

### DIFF
--- a/packages/fluxible-addons-react/FluxibleMixin.js
+++ b/packages/fluxible-addons-react/FluxibleMixin.js
@@ -117,7 +117,8 @@ var FluxibleMixin = {
         }
 
         if (!handler) {
-            throw new Error('storeListener attempted to add undefined handler. Make sure handlers actually exist.');
+            throw new Error('storeListener attempted to add undefined handler. Make sure handlers actually exist.' +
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

samlecuyer:
looks like the checks passed

samlecuyer:
@mridgway is this merge-able?

mridgway:
I think this error message may cause confusion for someone that is using a map of listeners instead of the default `onChange`. If we moved the warning to https://github.com/samlecuyer/fluxible/blob/81b3038a01db47bd6a0e0464876b6da6e945c542/packages/fluxible-addons-react/FluxibleMixin.js#L137 then we would be able to be very specific with which store is missing the handler and also use the `handler` string (if it is a string) in the warning.

redonkulus:
@samlecuyer thoughts on the feedback?

samlecuyer:
@redonkulus sorry, I was OOO last week.  I can open a new PR. I think that makes sense.

Issue #448, opened 2016-06-11
How to disable onScroll replaceState ?


Diokuz:
IOS safari has a bug https://forums.developer.apple.com/thread/36650
And we have a problem, because each scroll leads to history.replaceState.

How to disable it? We dont need scroll handling at all.

Maybe just make some `if (options.enableScroll)` here:

https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/handleHistory.js

?

mridgway:
A refactor for scroll state should be discussed in #336. Since this specific issue is resolved in #450 for now, I will close this issue.

PR #449, opened 2016-06-11
disable onScroll changeState

Diokuz:
Problem: https://github.com/yahoo/fluxible/issues/448
(SPA breaks in iOS safari after scrolling)

Solution: disable any scroll management in browser history if `enableScroll` is `false`.

### DIFF
--- a/packages/fluxible-router/lib/handleHistory.js
+++ b/packages/fluxible-router/lib/handleHistory.js
@@ -131,7 +131,9 @@ function createComponent(Component, opts) {
                 }
             }
 
-            window.addEventListener('scroll', this._onScroll);
+            if (oprions.enableScroll) {
+                window.addEventListener('scroll', this._onScroll);
+            }
         },
         _onScroll: function (e) {
             if (this._scrollTimer) {
@@ -173,12 +175,15 @@ function createComponent(Component, opts) {
             var navQuery = nav.query || {};
             var historyState = {
                 query: navQuery,
-                params: navParams,
-                scroll: {
+                params: navParams
+            };
+
+            if (oprions.enableScroll) {
+                historyState.scroll = {
                     x: window.scrollX || window.pageXOffset,
                     y: window.scrollY || window.pageYOffset
-                }
-            };
+                };
+            }
 
             var pageTitle = navParams.pageTitle || null;
 
@@ -214,7 +219,9 @@ function createComponent(Component, opts) {
         componentWillUnmount: function () {
             this._history.off(this._onHistoryChange);
 
-            window.removeEventListener('scroll', this._onScroll);
+            if (oprions.enableScroll) {
+                window.removeEventListener('scroll', this._onScroll);
+            }
 
             historyCreated = false;
         },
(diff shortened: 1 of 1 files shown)


yahoocla:

yahoocla:
CLA is valid!

Diokuz:
I can fix tests. Does pullrequest makes sense for you?

mridgway:
Some users may be relying on the scroll state to be present even if `enableScroll` is disabled (basically not having managed scroll positioning, but doing their own management). Maybe we should instead consider storing the scroll position in a store rather than in the history state.

@lingyan Thoughts?

lingyan:
@Diokuz Thanks for your contribution!  There are apps that currently rely on the scroll position saved in history state even when `enableScroll` is set to `false`.  So I'm afraid using `enableScroll` for controlling this will break them.

@mridgway Using store is an interesting idea.  We will need to associate the scroll position with each state.  We can do that easily by having a unique id for each state, and saving scroll state with the id in the store. The tricky thing is the backward compatibility: we need to figure out how not to break apps that currently are getting the scroll position for the state.  For next major version, we can have a better API to expose the scroll position, that would be a good way to go.  (Actually, we also talked about removing scroll related code completely out of fluxible-routr, and making it a plugin if needed.)

The reason the scroll position was saved to history state on scroll events was because of popstate triggered navigation.  By the time `handleHistory` gets the popstate event, the state has already updated.

Last alternative would be to introduce a new boolean config like `saveScrollPositionOnScroll` to control this (i.e. do not save scroll position).  It is not the best solution, but it is less change and easy to keep backward compatibility.

Let me know what you guys think.  Thanks.

Diokuz:
The main problem here is that Fluxible apps already broken in iOs Safari... because of too many replaceState. And I trying to find a way to fix it.
Looks like there is no hope for Apple to fix that. They sure it is okay: only 100 change/replaceState per session.

> Last alternative would be to introduce a new boolean config like saveScrollPositionOnScroll to control this (i.e. do not save scroll position). It is not the best solution, but it is less change and easy to keep backward compatibility.

I can do that. If you plan to drop this scroll-feature in next major release anyway, maybe that is a best option.

What do you say?

lingyan:
@Diokuz Just realized we are currently on 1.0.0 alpha on master branch for fluxible-router.  So we can break the backward compatibility, store the scroll position in a store and add a new function to expose the scroll position.   Which version of fluxible-router are you currently using?

Diokuz:
@lingyan `"fluxible-router": "0.4.10"` `"fluxible": "1.0.3"`, why?

> store the scroll position in a store and add a new function to expose the scroll position.

Can I just fix tests in this pr? Or all you say must be done in one pr?

lingyan:
@Diokuz The master branch is already on the 1.0 version, but only in alpha.  And you are using 0.4.x in your app.  You will need to create a new PR based off our https://github.com/yahoo/fluxible/tree/fluxible-router-0.4.x branch.

Adding a new config for not saving scroll data into history state (`saveScrollInState`) is fine with me for 0.4.x.  And our team (or you, or anyone that wants to contribute) can figure out a better approach for 1.x release.

@mridgway Your thoughts?

mridgway:
@lingyan That sounds reasonable to me.

Diokuz:
> You will need to create a new PR based off our https://github.com/yahoo/fluxible/tree/fluxible-router-0.4.x branch.
> Adding a new config for not saving scroll data into history state (saveScrollInState) is fine with me for 0.4.x. And our team (or you, or anyone that wants to contribute) can figure out a better approach for 1.x release.

@lingyan @mridgway https://github.com/yahoo/fluxible/pull/450

PR #451, opened 2016-06-17
[dispatchr] Defer rehydration of stores to when they are instantiated

mridgway:
This PR defers rehydration of stores until they are instantiated in `getStore()`. `getStore` is generally called from an action, dispatch, or from a view.  This could improve startup time for applications where stores are dehydrated but may not be immediately used for the initial render.

I changed the test to use the public interface for accessing the store rather than accessing the internal properties of dispatchr. The public interface has no breaking changes.


yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

Vijar:
👍

redonkulus:
👍

lingyan:
🚢

Issue #444, opened 2016-05-20
callAction has a bad condition


seeden:
Hi

I found a very strange behaviour. My done callback was very often called twice. The problem is on the next line https://github.com/yahoo/fluxible/blob/38630056bc52b430b64df26202be2d187990d729/packages/fluxible/utils/callAction.js#L29

I am using babel and babel will transform next action
```js
export function setAction(context, payload = {}, done) {
}
```

into the 
```js
export function setAction(context) {
  var payload = ....
}
```

As you can see transformed action named setAction has just one parameter (original has 3 parameters) and your condition action.length < 3 will be equals true. I think that too many of us are using babel in this moment and can have a same problem.

mridgway:
That's odd. Looks like this is the behavior of functions that use default parameters. Let me look in to whether this is new behavior from babel and whether we could treat this as a bug there, since it seems like babel shouldn't be changing the length.

mridgway:
Opened https://phabricator.babeljs.io/T7377 to follow up with Babel team.

mridgway:
After further investigation, this may be how ES6 spec works with default parameter values. We may have to find another solution for this internally.

seeden:
yes, the current condition is not good

mridgway:
So the only way to mitigate this that I can see is to check the return value against `undefined`. I'm not sure if there are valid cases where `undefined` could be the resolved value of the action's `Promise`. Seems like this would be a breaking change though either way.

mridgway:
After finding a bug internally that was related to this `arguments.length`, I think the best way forward is to remove the automatic resolution of Promises by returning values from actions. Instead, we would require Promises to be explicitly returned from the action or we will wait for the `done` callback to be called.

Anyone using the return value as the resolved value would need to use `return Promise.resolve(value)` instead or just returning the value from the action.

mridgway:
This would basically be just removing that line linked to in the origin comment (https://github.com/yahoo/fluxible/blob/master/packages/fluxible/utils/callAction.js#L29).

carystanley:
We do have actions without the callback param, that do not use promises, this is functionality that has been in place from the beginning. These actions are mostly doing synchronous logic and/or context.dispatch calls that would not really track callback, we have quite a few of these actions, so any breaking change introduced that removes support for this case would need substantial changes.

(just want you to be aware)

PR #460, opened 2016-07-18
Fix to devtools-plugin to support Actions without a "callback" param

carystanley:
@lingyan @redonkulus 

This should resolve the issue where calling executeAction on actions that do not contain a "callback" param while using devtools-plugin causes the request to hang.

### DIFF
--- a/packages/fluxible-plugin-devtools/src/lib/devtools-plugin.js
+++ b/packages/fluxible-plugin-devtools/src/lib/devtools-plugin.js
@@ -144,6 +144,10 @@ export default function devToolsPlugin() {
                         return action(ctx, payload, cb);
                     }
 
+                    function timedActionNoCallback(ctx, payload) {
+                        return timedAction(ctx, payload);
+                    }
+
                     function timedCallback(err, data) {
                         const endTime = Date.now();
                         const dur = endTime - actionReference.startTime;
@@ -155,7 +159,7 @@ export default function devToolsPlugin() {
 
                     return {
                         actionContext: actionContext,
-                        action: timedAction,
+                        action: (action.length < 3) ? timedActionNoCallback : timedAction,
                         payload: options.payload,
                         done: timedCallback
                     };


yahoocla:
CLA is valid!

lingyan:
@mridgway Any thoughts about this?

packages/fluxible-plugin-devtools/src/lib/devtools-plugin.js
```diff
@@ -144,6 +144,10 @@ export default function devToolsPlugin() {
                         return action(ctx, payload, cb);
                     }
 
+                    function timedActionNoCallback(ctx, payload) {
+                        return timedAction(ctx, payload);
```
Vijar: You might want to call these two lines before returning `timedAction`

```js
actionReference.startTime = Date.now();
actionReference.payload = JSON.stringify(payload);
```

Without a start time these actions might not end up in the correct place in the timeline. The payload will also be useful for debugging actions called w/o a callback.

carystanley: The call to "timedAction" sets those values already

carystanley: It seemed to show up in the correct place in the timeline for me

packages/fluxible-plugin-devtools/src/lib/devtools-plugin.js
```diff
@@ -155,7 +159,7 @@ export default function devToolsPlugin() {
 
                     return {
                         actionContext: actionContext,
-                        action: timedAction,
+                        action: (action.length < 3) ? timedActionNoCallback : timedAction,
```
redonkulus: Actions can also return promises so this might effect those too.

carystanley: timedAction does a return on the wrapped action, and I do a return on timedAction, so if a promise is returned it would be returned

packages/fluxible-plugin-devtools/src/lib/devtools-plugin.js
Vijar: Ah, you're right, I thought you were directly calling the action just w/o a callback.

carystanley:
Current behavior is that request crashes/hangs for actions that don't implement the "callback" param and use devtools-plugin, so this would be a big improvement.

mridgway:
Makes sense to me for now 👍 

I think some of this logic may change as a result of https://github.com/yahoo/fluxible/issues/444

lingyan:
@carystanley fluxible-plugin-devtools-0.1.5 released with this fix.  Thanks for your contrib!

PR #462, opened 2016-07-29
do not mutate global xhr path

kaesonho:
@lingyan @redonkulus @MattHo 

fix potential issue that if we are integrating `getXhrPath`, and generate different xhr path for each route, might cause some issue because we are always updating the global [xhrPath](https://github.com/yahoo/fluxible/pull/462/files#diff-b89bb231ea3f50b0af908412e103cc06R32) value. 
Which means we might serve xhrPath for previous route if somehow [Line 58](https://github.com/yahoo/fluxible/pull/462/files#diff-b89bb231ea3f50b0af908412e103cc06R58) is not executed or it's override by next request,  it's not clear but we got some issue on production if the traffic is getting heavier


yahoocla:
CLA is valid!

kaesonho:
@mridgway

mridgway:
👍

redonkulus:
:+1:

MattHo:
:+1:

lingyan:
Nice catch. @kaesonho 👍

PR #468, opened 2016-08-11
store static routes

kaesonho:
@redonkulus @lingyan @mridgway 

not only store router, also store static routes.

so that when we use `RouteStore.withStaticRoutes` and get `RECEIVE_ROUTES`, we will merge them with static routes, 

now static routes will be removed once we `RECEIVE_ROUTES` and it's not dehydrated

### DIFF
--- a/packages/fluxible-router/lib/RouteStore.js
+++ b/packages/fluxible-router/lib/RouteStore.js
@@ -158,6 +158,7 @@ RouteStore.withStaticRoutes = function (staticRoutes) {
     var staticRouter = new Router(staticRoutes);
     function StaticRouteStore() {
         RouteStore.apply(this, arguments);
+        this._routes = staticRoutes;
         this._router = staticRouter;
     }
     inherits(StaticRouteStore, RouteStore);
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

mridgway:
Static routes should not be removed when `RECEIVE_ROUTES` happens. They are merged together into a new object, but the static routes should not need to be dehydrated since they are static, they should be bundled with the rest of the JS.

lingyan:
@kaesonho `getRoutes()` does a shallow merge of the static routes and this._routes:https://github.com/yahoo/fluxible/pull/468/files#diff-f43def5c90ffb6949ffe453b37c2650eR169

kaesonho:
yeah I was checking 0.4.x, and make a PR with 1.x first, let me check, we probably just need this kind of fix on 0.4.x

mridgway:
I don't think that logic changed in 1.x. Can you help me understand the issue you are facing? Is it that you want to dehydrate static routes? That will break a lot of things I'm afraid.

kaesonho:
oh I didn't notice the [getRoutes](https://github.com/yahoo/fluxible/pull/468/files#diff-f43def5c90ffb6949ffe453b37c2650eR169) is doing a shallow merge here, 

so the problem is we are doing withStaticRoute only on server side, and it doesn't apply to client side, so I don't think we need a fix on fluxible-router, thanks for checking! @lingyan @mridgway

PR #470, opened 2016-08-29
Fix markdown syntax [ci skip]

ykzts:


yahoocla:
CLA is valid!

redonkulus:
Thanks!

PR #478, opened 2016-09-13
move href checking logic into isRoutable() to simplify dispatchNavAction()

lingyan:
* Refactor `dispatchNavAction()` for better readability
* Group href checking logic into `isRoutable()` so that it can be reused by custom NavLink classes created using `createNavLinkComponent`.

@redonkulus @kaesonho

### DIFF
--- a/packages/fluxible-router/tests/unit/lib/NavLink-test.js
+++ b/packages/fluxible-router/tests/unit/lib/NavLink-test.js
@@ -399,7 +399,6 @@ describe('NavLink', function () {
                 </MockAppComponent>
             );
             ReactTestUtils.Simulate.click(ReactDOM.findDOMNode(link), {button: 0});
-            ReactTestUtils.Simulate.click(ReactDOM.findDOMNode(link), {button: 0});
             window.setTimeout(function () {
                 expect(testResult.dispatch).to.equal(undefined);
                 done();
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

packages/fluxible-router/tests/unit/lib/NavLink-test.js
```diff
@@ -435,6 +434,20 @@ describe('NavLink', function () {
...
+                </MockAppComponent>
+            );
+            ReactTestUtils.Simulate.click(ReactDOM.findDOMNode(link), {button: 0});
+            window.setTimeout(function () {
+                expect(testResult.dispatch).to.equal(undefined);
+                expect(mockContext.executeActionCalls.length).to.equal(0);
+                done();
+            }, 10);
+        });
+
```
lingyan: Add missing unit test for url that does not match any route on client side

kaesonho:
for client only functions, do we want to add some defensive code? so that it won't be used incorrectly on server side.

other than this 👍

lingyan:
sure. I will add a check and throw error. @kaesonho

lingyan:
@kaesonho updated.

kaesonho:
👍

redonkulus:
👍

PR #492, opened 2016-09-28
Attempt to register stores that are not registered via getStore

redonkulus:
@mridgway @lingyan 

This will try to register a store via `getStore` if the constructor is passed in. This helps in cases where all stores may not be registered at init time.

### DIFF
--- a/packages/dispatchr/lib/DispatcherContext.js
+++ b/packages/dispatchr/lib/DispatcherContext.js
@@ -37,7 +37,11 @@ DispatcherContext.prototype.getStore = function getStore(name) {
     if (!this.storeInstances[storeName]) {
         var Store = this.dispatcher.stores[storeName];
         if (!Store) {
-            throw new Error('Store ' + storeName + ' was not registered.');
+            if ('function' === typeof name) {
+                this.dispatcher.registerStore(name);
+            } else {
+                throw new Error('Store ' + storeName + ' was not registered.');
+            }
         }
         this.storeInstances[storeName] = new (this.dispatcher.stores[storeName])(this.dispatcherInterface);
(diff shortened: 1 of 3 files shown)


yahoocla:
CLA is valid!

mridgway:
The only concern I would have about this is hidden bugs where a Store was not registered and therefore did not receive an action. Currently, if you checked the store for a value, it would throw. With this PR, you would get the store with its default state, but you wouldn't know that it didn't receive an event that happened previously.

packages/dispatchr/lib/DispatcherContext.js
```diff
@@ -37,7 +37,11 @@ DispatcherContext.prototype.getStore = function getStore(name) {
     if (!this.storeInstances[storeName]) {
         var Store = this.dispatcher.stores[storeName];
         if (!Store) {
-            throw new Error('Store ' + storeName + ' was not registered.');
+            if ('function' === typeof name) {
```
kaesonho: do we have any env variable anywhere? if yes we might want to print an warning on dev mode

redonkulus: A warning that the store may miss some dispatches? I think it would still leave the app in a potentially bad state, plus it would spam the user.

kaesonho: yeah I mean 

ideally we should make stores registered, if there's any case that user getStore without register it, then there must be something they need to fix.

previously we let it throw, now if we handle this, people would never aware of the issue.

redonkulus:
Closing in favor of https://github.com/yahoo/fluxible/pull/496

PR #496, opened 2016-10-03
Allow error handler to be used instead of throwing

redonkulus:
@mridgway @lingyan @kaesonho 

Opening this up for discussion. 

We have the use case where we want to beacon back to the server when a store registration fails. We can't do this today because dispatchr [throws](https://github.com/yahoo/fluxible/blob/master/packages/dispatchr/lib/DispatcherContext.js#L40) with no way of handling it. In this PR, I've introduced a function handler that can be passed in so that applications can do something with the error (eventually throwing themselves).

Additionally, I've tried to make the error handler sane by passing in `type` and `meta` information arguments so applications can handle various errors differently.

This would remove the need for https://github.com/yahoo/fluxible/pull/492

### DIFF
--- a/packages/dispatchr/lib/DispatcherContext.js
+++ b/packages/dispatchr/lib/DispatcherContext.js
@@ -37,7 +37,18 @@ DispatcherContext.prototype.getStore = function getStore(name) {
     if (!this.storeInstances[storeName]) {
         var Store = this.dispatcher.stores[storeName];
         if (!Store) {
-            throw new Error('Store ' + storeName + ' was not registered.');
+            var message = 'Store ' + storeName + ' was not registered.';
+            if (this.dispatcher.errorHandler) {
+                this.dispatcher.errorHandler(this.context, {
+                    type: 'STORE_UNREGISTERED',
+                    meta: {
+                        storeName: storeName,
+                        message: message
+                    }
+                });
+            } else {
+                throw new Error(message);
+            }
         }
         this.storeInstances[storeName] = new (this.dispatcher.stores[storeName])(this.dispatcherInterface);
         if (this.rehydratedStoreState && this.rehydratedStoreState[storeName]) {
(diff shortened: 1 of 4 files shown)


yahoocla:
CLA is valid!

packages/dispatchr/lib/DispatcherContext.js
```diff
@@ -14,6 +14,7 @@ var debug = require('debug')('Dispatchr:DispatcherContext');
  * @constructor
  */
 function DispatcherContext (dispatcher, context) {
+    this.context = context;
```
redonkulus: We had no way of getting to `context` before, so I added it here so error handlers can use it later.

packages/dispatchr/lib/DispatcherContext.js
```diff
@@ -37,7 +38,18 @@ DispatcherContext.prototype.getStore = function getStore(name) {
     if (!this.storeInstances[storeName]) {
         var Store = this.dispatcher.stores[storeName];
         if (!Store) {
-            throw new Error('Store ' + storeName + ' was not registered.');
+            var message = 'Store ' + storeName + ' was not registered.';
+            if (this.dispatcher.errorHandler) {
+                this.dispatcher.errorHandler(this.context, {
```
redonkulus: something like this could be leveraged across various areas that `throw` today.

lingyan:
Overall approach is reasonable to me.  `errorHandler` API can be discussed.

redonkulus:
@mridgway @lingyan do you have a better name for the API?

redonkulus:
@lingyan @mridgway @kaesonho I've added tests and docs, please review. I'm sticking with `errorHandler` as the name unless a better one is suggested.

lingyan:
I'm fine with `errorHandler` name.  I was originally thinking about the parameters for the `errorHandler` can be an Error object.  But the current implementation in this PR is also fine to me.

kaesonho:
👍

PR #499, opened 2016-10-13
Updating connectToStores to account for React Transition Group methods

realseanp:
When you have a child of React Transition Group that is wrapped in the connectToStores higher order component the Transition Group lifecycle methods (https://facebook.github.io/react/docs/animation.html#low-level-api-reacttransitiongroup) will be stripped from the component. The fix for this is to account for the Transition Group lifecycle methods in connectToStores.


yahoocla:

yahoocla:
CLA is valid!

redonkulus:
Closing based on https://github.com/yahoo/fluxible/issues/500

PR #502, opened 2016-10-18
Allow custom props to be added to filtered ones

redonkulus:
@kaesonho @lingyan @mridgway 

This basically allows consumers to [pass in additional properties](https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/createNavLinkComponent.js#L291) to be filtered before the render call. The consumer will return an array from `getFilteredProps` like so:

```js
createNavLinkComponent({
    // ...
    getFilteredProps: function () {
        return [
            'bindClickEvent',
            'follow',
            'i13nModel',
            'isLeafNode',
            'model',
            'scanLinks'
        ];
    },
    // ...
});
```

Once approved I will add this to the 1.x branch.

### DIFF
--- a/packages/fluxible-router/lib/createNavLinkComponent.js
+++ b/packages/fluxible-router/lib/createNavLinkComponent.js
@@ -141,6 +141,15 @@ module.exports = function createNavLinkComponent (overwriteSpec) {
         getDefaultChildProps: function () {
             return {};
         },
+        /**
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

lingyan:
👍

kaesonho:
👍

PR #505, opened 2016-11-16
For invalid href and route name, only throw error in development env,…

lingyan:
… render with empty href in prod with console error

Avoid crashing node process/ui thread in production just because one NavLink is configured wrong.

Will backport to 0.4.x after this PR is reviewed.

@kaesonho @mridgway @redonkulus

### DIFF
--- a/packages/fluxible-router/lib/createNavLinkComponent.js
+++ b/packages/fluxible-router/lib/createNavLinkComponent.js
@@ -1,13 +1,14 @@
 /**
- * Copyright 2015, Yahoo! Inc.
+ * Copyright 2015-Present, Yahoo! Inc.
  * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
-/*global window */
+/*global window,process */
 'use strict';
 var React = require('react');
 var RouteStore = require('./RouteStore');
 var debug = require('debug')('NavLink');
 var navigateAction = require('./navigateAction');
+var __DEV__ = process.env.NODE_ENV !== 'production';
 
 function objectWithoutProperties(obj, keys) {
     var target = {};
@@ -152,11 +153,6 @@ module.exports = function createNavLinkComponent (overwriteSpec) {
             if (!href && routeName) {
                 href = routeStore.makePath(routeName, navParams, queryParams);
             }
-            if (!href) {
-                throw new Error('NavLink created without href or unresolvable ' +
-                    'routeName \'' + routeName + '\' with params ' +
-                    JSON.stringify(navParams));
-            }
             return href;
         },
         /**
@@ -292,12 +288,23 @@ module.exports = function createNavLinkComponent (overwriteSpec) {
             this.dispatchNavAction(e);
         },
         render: function () {
-            var href = this._getHrefFromProps(this.props);
-            var activeClass = this.props.activeClass;
-            var activeStyle = this.props.activeStyle;
-            var activeElement = this.props.activeElement;
+            var props = this.props;
+            var href = this._getHrefFromProps(props);
+            if (!href) {
+                if (__DEV__) {
+                    throw new Error('NavLink created with invalid href \'' + props.href +
+                        '\'or unresolvable routeName \'' + props.routeName);
+                } else {
+                    console.error('Error: Invalid NavLink, skip rendering...', props);
+                    return null;
+                }
+            }
 
-            var childProps = objectWithoutProperties(this.props, [
+            var activeClass = props.activeClass;
+            var activeStyle = props.activeStyle;
+            var activeElement = props.activeElement;
+
+            var childProps = objectWithoutProperties(props, [
                 'activeClass',
                 'activeElement',
                 'activeStyle',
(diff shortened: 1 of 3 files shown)


yahoocla:
CLA is valid!

packages/fluxible-router/tests/unit/lib/NavLink-test.js
```diff
@@ -3,19 +3,23 @@
...
+
+var jsdom = require('jsdom');
+var expect = require('chai').expect;
+var fs = require('fs');
+var resolve = require('resolve');
+var createMockComponentContext = require('fluxible/utils/createMockComponentContext');
+var navigateAction = require('../../../').navigateAction;
+var RouteStore = require('../../../').RouteStore;
+var ORIG_NODE_ENV = process.env.NODE_ENV;
+
```
lingyan: OCD reorg

packages/fluxible-router/tests/unit/lib/NavLink-test.js
```diff
@@ -25,50 +29,68 @@ var TestRouteStore = RouteStore.withStaticRoutes({
...
+        }
+    });
+}
+
+function tearDown() {
+    delete global.window;
+    delete global.document;
+    delete global.navigator;
+    process.env.NODE_ENV = ORIG_NODE_ENV;
+}
```
lingyan: refactor the logic in old `beforeEach` to be able to share between 3 test suites.

packages/fluxible-router/tests/unit/lib/NavLink-test.js
```diff
@@ -111,16 +133,6 @@ describe('NavLink', function () {
...
-        it('should throw if href and routeName undefined', function () {
-            var navParams = {a: 1, b: 2};
-            expect(function () {
-                ReactTestUtils.renderIntoDocument(
-                    <MockAppComponent context={mockContext}>
-                        <NavLink navParams={navParams} />
-                    </MockAppComponent>
-                );
-            }).to['throw']();
-        });
```
lingyan: moved out into 2 other test suites (one for dev env, the other for prod env) to test diff behaviors in diff environments.

packages/fluxible-router/lib/createNavLinkComponent.js
```diff
@@ -1,13 +1,14 @@
...
  * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
-/*global window */
+/*global window,process */
 'use strict';
 var React = require('react');
 var RouteStore = require('./RouteStore');
 var debug = require('debug')('NavLink');
 var navigateAction = require('./navigateAction');
+var __DEV__ = process.env.NODE_ENV !== 'production';
```
mridgway: We have typically inlined this and used webpack to replace `process.env.NODE_ENV`.

mridgway:
I've always been a fan of throwing for developer errors, but I'm don't feel strongly enough to reject this.

lingyan:
@mridgway I hear you.  We had a production outage this morning because one of the link (out of many links) on the page failed to render.  And it crashed the node processes.  That is essentially what triggered this PR.   We could ask NavLink users to validate href/routeName ahead of using NavLink.  But a little hard to scale as every NavLink has to run the validation check :(

I'm open for better suggestions for sure.

packages/fluxible-router/lib/createNavLinkComponent.js
lingyan: ha.  Stole it from [fluxible](https://github.com/yahoo/fluxible/blob/master/packages/fluxible/lib/Fluxible.js#L11).  Dont think it is inlined there.

webpack will replace here as well.  And I thought we had concerns about accessing `process.env.NODE_ENV` on every invocation of a hot function (like NavLink render).  Maybe that is no longer a concern?

packages/fluxible-router/lib/createNavLinkComponent.js
```diff
@@ -292,12 +289,24 @@ module.exports = function createNavLinkComponent (overwriteSpec) {
...
+            var props = this.props;
+            var href = this._getHrefFromProps(props);
+            if (!href) {
+                if (__DEV__) {
+                    throw new Error('NavLink created with invalid href \'' + props.href +
+                        '\'or unresolvable routeName \'' + props.routeName);
+                } else {
+                    var logError = (this.context.logger && this.context.logger.error) || console.error;
+                    logError('Error: Invalid NavLink, skip rendering...', props);
+                    return null;
```
kaesonho: instead of returning `null`, I'm thinking if we should just render empty href, and when user click, should return  https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/createNavLinkComponent.js#L262 and fallback to native anchor behavior with empty href.

lingyan: I like that. Updating. @kaesonho

lingyan: Done. @kaesonho Check it out :)

kaesonho:
👍

lingyan:
@mridgway @redonkulus  more thoughts on this PR?  I updated the PR to render with provided href prop in prod env (in the error case), instead of rendering null.

redonkulus:
👍

PR #509, opened 2016-12-01
For invalid href and route name, only throw error in development env,…

lingyan:
… render with empty href in prod with console error

backport the fix from https://github.com/yahoo/fluxible/pull/505

@kaesonho @redonkulus @mridgway


yahoocla:
CLA is valid!

redonkulus:
:+1:

PR #510, opened 2016-12-07
Expose WrappedComponent from StoreConnector

kinetifex:
This allows easy access to the underlying component for tests as in `react-intl`:
https://github.com/yahoo/react-intl/blob/master/src/inject.js#L32
https://github.com/yahoo/react-intl/wiki/Testing-with-React-Intl

### DIFF
--- a/packages/fluxible-addons-react/connectToStores.js
+++ b/packages/fluxible-addons-react/connectToStores.js
@@ -59,6 +59,7 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
     });
 
     hoistNonReactStatics(StoreConnector, Component);
+    StoreConnector.WrappedComponent = Component;
 
     return StoreConnector;
 }
--- a/packages/fluxible-router/lib/handleHistory.js
+++ b/packages/fluxible-router/lib/handleHistory.js
@@ -286,7 +286,7 @@ function createComponent(Component, opts) {
     // Copy statics to HistoryHandler
     hoistNonReactStatics(HistoryHandler, Component);
 
-    HistoryHandler.wrappedComponent = Component;
+    HistoryHandler.wrappedComponent = HistoryHandler.WrappedComponent = Component;
 
     return handleRoute(HistoryHandler);
 }
--- a/packages/fluxible-router/lib/handleRoute.js
+++ b/packages/fluxible-router/lib/handleRoute.js
@@ -52,7 +52,7 @@ function createComponent(Component) {
     // Copy statics to RouteHandler
     hoistNonReactStatics(RouteHandler, Component);
 
-    RouteHandler.wrappedComponent = Component;
+    RouteHandler.wrappedComponent = RouteHandler.WrappedComponent = Component;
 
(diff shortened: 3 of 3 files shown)


yahoocla:

yahoocla:
CLA is valid!

packages/fluxible-addons-react/connectToStores.js
```diff
@@ -59,6 +59,7 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
     });
 
     hoistNonReactStatics(StoreConnector, Component);
+    StoreConnector.WrappedComponent = Component;
```
mridgway: In our other Fluxible related components we have used the lowercase `wrappedComponent` so we may want to be consistent here. I'm curious what other libraries are using in case there is a standard.

kinetifex: So I went with capital `WrappedComponent` to align with how [react-intl exposes it](https://github.com/yahoo/react-intl/blob/master/src/inject.js#L32) as well as how [react-router exposes it](https://github.com/reactjs/react-redux/blob/master/docs/api.md#static-properties). However, I see now that the fluxible-router package in this repo is using `wrappedComponent` as you mentioned.

It is a good call out. Does the fluxible team want to align on lowercased within the fluxible project or shift to uppercase in [RouteHandler](https://github.com/yahoo/fluxible/blob/44f39fe1e917ba427b115ec78268c0a18acd8eb7/packages/fluxible-router/lib/handleRoute.js#L55) and [HistoryHandler](https://github.com/yahoo/fluxible/blob/51330c95fab507d47f445ad7d5bf85928fed096f/packages/fluxible-router/lib/handleHistory.js#L289)?

mridgway: Thanks for taking a look around at other libraries. I agree we should try to align with the rest of the community. Let's keep this PR the way it is.

For now, maybe we attach both `wrappedComponent` and `WrappedComponent` in the other files to maintain backwards compatibility but provide a way forwards.  Then we'll take a look at removing the lowercase version in our other HoCs in the next breaking versions. 

Not ideal, but at least better aligned.

kinetifex: When do you expect this to be merged/published? Did you want me to the attaching of `WrappedComponent` in the other files?

redonkulus: @kinetifex yes let's update the other files too to be consistent.

redonkulus:
@kinetifex can you support `wrappedComponent` and `WrappedComponent`? Otherwise your changes will break people down stream. In the next major release of those packages we can remove the old one.

kinetifex:
@redonkulus Ok, now supporting both `wrappedComponent` and `WrappedComponent` for existing cases.

redonkulus:
Ok code looks good. Thanks!

redonkulus:
@kinetifex these changes have been published in latest fluxible-router and fluxible-addons-react packages.

PR #516, opened 2017-02-14
skip updating scroll when it is not changed

kaesonho:
@lingyan @redonkulus 

on scroll we keep using replaceState to update scroll position, while this causes an unexpected behavior on IOS Safari with universal links bar showing.

The issue is when it shows and we try to replaceState, browser will try to reset the scroll position to 0 and hide the bar. 

The solution here is don't update state when scroll position remain unchanged. It's unused anyways and we should just prevent duplicated action.

### DIFF
--- a/packages/fluxible-router/lib/handleHistory.js
+++ b/packages/fluxible-router/lib/handleHistory.js
@@ -210,9 +210,17 @@ function createComponent(Component, opts) {
         },
         _saveScrollPosition: function (e) {
             var historyState = (this._history.getState && this._history.getState()) || {};
+            var scrollX = window.scrollX || window.pageXOffset;
+            var scrollY = window.scrollY || window.pageYOffset;
+            // reduce unused replaceState
+            // also prevent IOS Safari reset scroll position to 0 with universal link bar showing
+            if (historyState.scroll && historyState.scroll.x === scrollX && historyState.scroll.y === scrollY) {
+                debug('skip updating scrolling position with same position', historyState.scroll);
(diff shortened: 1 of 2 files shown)


packages/fluxible-router/lib/handleHistory.js
```diff
@@ -210,9 +210,17 @@ function createComponent(Component, opts) {
         },
         _saveScrollPosition: function (e) {
             var historyState = (this._history.getState && this._history.getState()) || {};
+            var scrollX = window.scrollX || window.pageXOffset;
+            var scrollY = window.scrollY || window.pageYOffset;
+            // reduce unused replaceState
+            // also prevent IOS Safari reset scroll position to 0 with universal link bar showing
+            if (historyState.scroll && historyState.scroll.x === scrollX && historyState.scroll.y === scrollY) {
```
lingyan: I had a thought about pushing the comparison down the `_history.replaceState()`, which would be more generic.  But that could make the replaceState() impl a little more complex.  I'm good with this approach, which _history.replaceState() caller compares the parameters (its trying to update) explicitly.  👍

lingyan:
👍

packages/fluxible-router/lib/handleHistory.js
kaesonho: I actually didn't aware that we have a wrapper for `history`, to me it makes sense to implement this logic in `_history.replaceState()` .

redonkulus:
👍

PR #495, opened 2016-09-30
[resolves #494] Remove setImmediate from rehydration promise

kirbdee:
For https://github.com/yahoo/fluxible/issues/494

Issue #494, opened 2016-09-30
setImmediate forces Render to be async


kirbdee:
I'm observing the issue where the time between hydration and rendering/mounting can increase / blocked due to setImmediate forcing rendering to be async and putting it back into the "task queue"

My observations: https://github.com/kirbdee/FluxibleSample/blob/master/README.md
Code: https://github.com/kirbdee/FluxibleSample

mridgway:
I'm curious why removing `setImmediate` actually fixes the issue since `.then`'s callback is also guaranteed to be async and should push the render back as well. The ordering of these methods could be browser specific which is something look into.

WRT the mention of handling errors, the `setImmediate` is there so that handles in the callback function do not get caught by an internal Promise within Fluxible.

```js
app.rehydrate(state, (err, context) => {
   // err can be handled here
  if (err) {
      handleError(err);
      return;
  }

  myUndefinedFunction() // <-- throws an error, caught by internal Fluxible promise silently
});
```

This is pretty critical in my opinion, because it seems like something went wrong with Fluxible even though there was an error in the user's code. Unless we can solve this, I don't think we'd be willing to remove the `setImmediate`.

I'm curious about your example you mention that the second JS file is critical, but you want it to be loaded after the initial execution of the first file which seems contradictory to me. I would have suggested `async` as well, so I'm not sure I fully understand your reasoning there.

kirbdee:
Regarding the error handling, I see! but in Promise(we're using es6-promise right?) couldn't we do this:

app.rehydrate(state, (err, context) => {
   // err can be handled here
  if (err) {
      handleError(err);
      return;
  }

  myUndefinedFunction() // <-- throws an error, caught by internal Fluxible promise silently
}).catch(function(error) {
  throw error //<-- thrown back up
});

per http://www.html5rocks.com/en/tutorials/es6/promises/#toc-error-handling (which was referred to from https://www.npmjs.com/package/es6-promise)

mridgway:
Yeah, that's totally possible, but creates confusion since you now have to deal with both Promises and callbacks. I really hate having to support both >.<

kirbdee:
@mridgway added another example internally instead

kirbdee:
setImmediate just seems so out of place to me, just to handle errors

PR #518, opened 2017-02-16
Update callAction() to use less setImmediates

ericf:
This removes two `setImmediate()` calls in the common action path and leverages promises more for enforcing async calling of actions.

Since Promises are a standardized EMCAScript feature, JS VMs are putting effort into improving their performance. While, currently, `setImmediate()` is a non-standard API.

What we're trying to optimize here is limiting the amount of time the code is waiting for turns of the JS event loop, while still keeping the async semantics. The idea is that by using promises the JS engine will be able to group async operations within the same micro-task queue, and therefore within the same turn of the event loop.

**Previous Async Turns (Promise-returning Action):**
```
└─┬ setImmediate()
  └─┬ Promise#then()
    └─┬ Promise#then()   (when callback is provided)
      └── setImmediate() (when callback is provided)
```

**Previous Async Turns (Callback-based Action):**
```
└─┬ setImmediate()
  └─┬ Promise#then()   (when callback is provided)
    └── setImmediate() (when callback is provided)
```

**New Async Turns (Promise-returning Action):**
```
└─┬ Promise#then()
  └── Promise#then() (when callback is provided)
```

**New Async Turns (Callback-based Action):**
```
└─┬ Promise#then()
  └── Promise#then() (when callback is provided)
```

**Note:** There's implicitly another `Promise#then()` for this new code, but JS engines should ideally be able to optimize settling this promise within the same turn of the event loop. This is why I didn't added it to the above since we're not explicitly calling `.then()` on the middle promise.

### DIFF
--- a/packages/fluxible/utils/callAction.js
+++ b/packages/fluxible/utils/callAction.js
@@ -14,38 +14,43 @@ require('setimmediate');
  * otherwise, Promise invocation.
  */
 function callAction (actionContext, action, payload, done) {
-    var executeActionPromise = new Promise(function (resolve, reject) {
-        setImmediate(function () {
-            try {
-                var syncResult = action(actionContext, payload, function (err, result) {
-                    if (err) {
-                        reject(err);
-                    } else {
-                        resolve(result);
-                    }
-                });
-                if (isPromise(syncResult)) {
-                    syncResult.then(resolve, reject);
-                } else if (action.length < 3) {
-                    resolve(syncResult);
+    // Use a promise to force the action to be called async
+    var executeActionPromise = Promise.resolve().then(function () {
+        // Assume the action returns a promise since it doesn't take a callback
+        if (action.length < 3) {
+            return action(actionContext, payload);
+        }
+
+        // Return a nested promise to wrap the action invocation so its result
+        // passed to its callback can be captured
+        return new Promise(function (resolve, reject) {
+            action(actionContext, payload, function (err, result) {
+                if (err) {
+                    reject(err);
+                } else {
+                    resolve(result);
                 }
-            } catch (e) {
-                reject(e);
-            }
+            });
         });
     });
 
     if (done) {
+        // Once our promise (and any nested promises) fully settle, call the
+        // `done()` callback provided to us and pass either the result or error
         executeActionPromise
-            .then(function(result) {
-                // Ensures that errors in callback are not swallowed by promise
-                setImmediate(done, null, result);
-            }, function (err) {
-                // Ensures that errors in callback are not swallowed by promise
-                setImmediate(done, err);
+            .then(function (result) {
+                done(null, result);
+            }, done)
+            ['catch'](function (err) {
+                // Ensures that thrown errors in the `done()` callback above are
+                // not swallowed by promise
+                setImmediate(function () {
+                    throw err;
+                });
             });
     }
 
+    // Finally return our original promise with the result of calling the action 
     return executeActionPromise;
 }
 


packages/fluxible/utils/callAction.js
```diff
@@ -14,38 +14,43 @@ require('setimmediate');
...
-                });
-                if (isPromise(syncResult)) {
-                    syncResult.then(resolve, reject);
-                } else if (action.length < 3) {
-                    resolve(syncResult);
+    // Use a promise to force the action to be called async
+    var executeActionPromise = Promise.resolve().then(function () {
+        // Assume the action returns a promise since it doesn't take a callback
+        if (action.length < 3) {
+            return action(actionContext, payload);
```
ericf: The action can be sync or return a promise here. Also if the action throws, it will be properly caught.

packages/fluxible/utils/callAction.js
```diff
@@ -14,38 +14,43 @@ require('setimmediate');
...
+        }
+
+        // Return a nested promise to wrap the action invocation so its result
+        // passed to its callback can be captured
+        return new Promise(function (resolve, reject) {
+            action(actionContext, payload, function (err, result) {
```
ericf: If the action throws, this nested promise will be rejected.

packages/fluxible/utils/callAction.js
```diff
@@ -14,38 +14,43 @@ require('setimmediate');
...
-                // Ensures that errors in callback are not swallowed by promise
-                setImmediate(done, err);
+            .then(function (result) {
+                done(null, result);
+            }, done)
+            ['catch'](function (err) {
+                // Ensures that thrown errors in the `done()` callback above are
+                // not swallowed by promise
+                setImmediate(function () {
+                    throw err;
```
ericf: Relegate the usage of `setImmediate()` only to the area where it makes a difference, capturing the errors thrown by a `done()` callback.

mridgway:
👍

redonkulus:
These changes look good to me. I'd like to incorporate the changes for the `rehydrate` method too, based on #495 findings from @kirbdee. Would also be good to have before and after perf numbers for reference.

mridgway:
I believe that #495's changes remove one of the "necessary" `setImmediate`s because anything after rehydrate that throws would be caught by those promises.

kirbdee:
for #495 you presented your concernt https://github.com/yahoo/fluxible/issues/494, I gave an example of rethrowing the error instead. there's also documentation there showing the benefit and this was showing a simple app. I don't think it hurts to maybe just try that PR in a bucket or something and see what performance benefits actually happen in RL

ericf:
@kirbdee the issue is that if the callback throws it'll be hard to debug in Node because the error can be swallowed by the promise. The fix for this is like what I have in this PR:

```js
somePromise
  .then(
    (result) => callback(null, result),
    (err) => callback(err)
  )
  .catch((err) => {
    setImmediate(() => {throw err;});
  });
```

This way the error will be logged in Node when using Domains.

kirbdee:
Ahh My pr doesn't reflect the disussion on the issue thread.

`app.rehydrate(state, (err, context) => {
   // err can be handled here
  if (err) {
      handleError(err);
      return;
  }

  myUndefinedFunction() // <-- throws an error, caught by internal Fluxible promise silently
}).catch(function(error) {
  throw error //<-- thrown back up
});`

Is what I was suggesting. Curious why you still need / want setImmediate

ericf:
> Is what I was suggesting. Curious why you still need / want setImmediate

Because it will be swallowed in older versions of Node (< 6.6.0) or when using Domains.

kirbdee:
is rehydrate run severside? from my understanding this is used client side

lingyan:
@kirbdee is right. rehydrate() is only used on client side.  If there is no error-being-swallowed issue on browser side, setImmediate won't be necessary for rehydrate().

ericf:
Yeah, in newer versions of Chrome or Safari the error won't be swallowed, but in any browser that needs to use a Promise polyfill it will.

lingyan:
In that case, it is safer to keep setImmediate in `catch` for rehydrate() then.  And adding setImmediate there does not really matter performance wise.

kirbdee:
Got it, i'm okay with having setImmediate in a catch just to rethrow. This should help with the rendering I was mentioning. I'll make this reflection in the PR if everyone is chill with that for rehydrate

lingyan:
@kirbdee Sounds good.  You can update your PR for the rehydrate() change, and this PR will just be for callAction changes.

packages/fluxible/utils/callAction.js
```diff
@@ -14,38 +14,43 @@ require('setimmediate');
...
+        // `done()` callback provided to us and pass either the result or error
         executeActionPromise
-            .then(function(result) {
-                // Ensures that errors in callback are not swallowed by promise
-                setImmediate(done, null, result);
-            }, function (err) {
-                // Ensures that errors in callback are not swallowed by promise
-                setImmediate(done, err);
+            .then(function (result) {
+                done(null, result);
```
ericf: I'm second-guessing myself that this is correct. I think we might need to keep the `setImmediate()` here in the result callback and really the only thing we can do is remove the first `setImmediate()`.

mridgway: It's different behavior for sure. You would no longer be able to handle exceptions within the context of the request. In theory, if an exception like this does occur, it's a critical bug in the code and recovering may not be safe, but it is a divergence from our original thinking of being able to handle it within the context of a request.

kirbdee: i like it! 👍  if you think this is valid for client side as well I'd like to adjust the rehydration stuff as well

kirbdee: Think this just for server though?

redonkulus:
Fixed via https://github.com/yahoo/fluxible/pull/537

PR #519, opened 2017-02-22
Throw correct 'store has no method' error

ash14:
My app was throwing an error "Cannot read property 'dispatcher' of undefined" which is misleading, since the correct error message should have been "Store does not have a method called [...]". This PR fixes that, with updated tests.


redonkulus:
@ash14 good find. Thanks for the fix.

redonkulus:
Available in dispatchr@1.1.1

PR #521, opened 2017-03-13
Update examples to React Router 3

pavestru:
Fluxible continues to work well with react-router 3.

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -19,7 +19,7 @@
     "history": "^1.17.0",
     "react": "0.14.x",
     "react-dom": "0.14.x",
-    "react-router": "^1.0.3",
+    "react-router": "^3.0.0",
     "serialize-javascript": "^1.0.0",
     "serve-favicon": "^2.1.6"
   },
--- a/examples/react-router/client.js
+++ b/examples/react-router/client.js
@@ -6,7 +6,6 @@
 'use strict';
 var React = require('react');
 var ReactDOM = require('react-dom');
-var createBrowserHistory = require('history/lib/createBrowserHistory');
 var debug = require('debug');
 var bootstrapDebug = debug('Example');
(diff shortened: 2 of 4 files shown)


redonkulus:
@pavestru why not React router 4?

pavestru:
Reasons why React Router 3 (and not React Router 4):
  - RR4 requires react v15 as a [peerDependency](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/package.json#L40). Fluxible examples still use react@0.14, so RR3 is just fine supporting both [react v0.14 and v15](https://github.com/ReactTraining/react-router/blob/v3/package.json#L43), thus suitable when upgrading the whole examples to react v15.
  - RR4 seems to be conceptually rather a new router library. Much bigger API change than between v1 and v3.
  - RR3+Fluxible has been production tested, while RR4 has been released out of beta just 3 days ago

Once the examples sub-repo updates react to v15, I see no problem in updating to react-router 4. They're having a [server rendering section](https://reacttraining.com/react-router/web/guides/server-rendering) in their docs. I recognize its shift to declarative approach to routing and in the meantime the [community's rage](https://news.ycombinator.com/item?id=12517962) about breaking API completely with every major version will calm down.

redonkulus:
Thanks for the info. The move to React 15 in our examples is only due to time constraints rather than  any technical issue. Its really just something we missed when we upgrade the rest of the code base to React 15.  I've opened [a separate issue](https://github.com/yahoo/fluxible/issues/523) to track that.

Other than that, these changes look fine to me.

redonkulus:
Thanks!

redonkulus:
@pavestru FYI I've updated the examples to use React 15, so react-router@4 would be testable now.

pavestru:
@redonkulus Great, I'll take a look.

PR #524, opened 2017-03-21
allow for dynamic initialState

pwmckenna:
This will allow us to set a date to "now" for instance, without it always being set to the moment the store is defined. Alternatively, we could add support for spec.initialize?

### DIFF
--- a/packages/fluxible-reducer-store/lib/createReducerStore.js
+++ b/packages/fluxible-reducer-store/lib/createReducerStore.js
@@ -14,7 +14,7 @@ module.exports = function createReducerStore(spec) {
         storeName: spec.storeName,
         handlers: {},
         initialize: function () {
-            this.state = spec.initialState;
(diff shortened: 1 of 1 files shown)


yahoocla:

mridgway:
I think having this function called from the constructor could create some odd behavior since the stores are lazily instantiated rather than all instantiated at the same time. It seems like what you would really want is a way to specify initial state at the time of the FluxibleContext creation that would eventually be passed in to the store constructor. I think this makes sense as a use case, we just have to figure out the best API to do so.

As a workaround you could dispatch the date immediately after `app.createContext` and have your store update based on that payload. I think this might be cleaner from an implementation standpoint for now, although it's a lot of extra work for initial state.

pwmckenna:
@mridgway thanks for the response. what you're suggesting is what we ended up doing, and given what you're saying about the lazy instantiation actually seems like a better approach. Don't worry about the use case on our behalf :)

Issue #525, opened 2017-03-27
[fluxible-router] replaceState always updates <title> on scroll


alexparish:
Is there a reason History.js is set up to always update the title on scroll?

It seems unperformant to update the DOM any time the user scrolls the page.

There could be a good reason for this but unfortunately I'm unfamiliar with the project.

redonkulus:
@alexparish looking at the code, where do you see this being added on scroll?

alexparish:
https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/handleHistory.js#L136

_saveScrollPosition is called and the title is finally updated here: https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/History.js#L141

redonkulus:
I see now, thanks. I agree that the title probably doesn't need to be updated each time. We could check if the title is the same before changing it or have some flag that prevents it from happening. Are you seeing performance issues because of this?

/cc @lingyan

alexparish:
I've seen scrolling is sometimes janky in our app but I can't definitively associate it to the title updates.

I think checking if the title is the same before updating would be great.

redonkulus:
@alexparish want to PR it?

alexparish:
Sure! Am on it.

redonkulus:
@alexparish this is now published in fluxible-router@1.0.1. Thanks again!

alexparish:
Awesome, thank you!

PR #526, opened 2017-03-27
Skip document title update when it is not changed

alexparish:
Fix for issue https://github.com/yahoo/fluxible/issues/525

### DIFF
--- a/packages/fluxible-router/lib/History.js
+++ b/packages/fluxible-router/lib/History.js
@@ -145,11 +145,11 @@ History.prototype = {
     },
 
     /**
-     * Sets document title. No-op if title is empty.
+     * Sets document title. No-op if title is empty or same as current document title.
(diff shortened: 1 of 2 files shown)


yahoocla:

redonkulus:
@alexparish thanks. Can you sign the CLA and add a unit test?

alexparish:
@redonkulus I've added unit tests to ensure the document title is updated when a new title is passed but I'm not sure how to test that a property is not assigned the same value (for the case the same title is passed). Easy enough to spy on a function call but not so easy with properties.

Do you know if this is even possible. Perhaps using sinon? If not feel free to accept this PR.

redonkulus:
@alexparish this looks good to me. Having a test that changing the title still works, is good to me. Thanks!

PR #530, opened 2017-04-13
Added create-react-class for React 15.5 support

J-Dickson:
Switched out React.createClass for create-react-class in fluxible-addons-react and fluxible-router due to API change in React v15.5.

### DIFF
--- a/packages/fluxible-addons-react/package.json
+++ b/packages/fluxible-addons-react/package.json
@@ -14,7 +14,8 @@
   },
   "dependencies": {
     "hoist-non-react-statics": "^1.0.4",
-    "inherits": "^2.0.1"
+    "inherits": "^2.0.1",
+    "create-react-class": "^15.5.1"
   },
   "peerDependencies": {
     "fluxible": ">=1.0.0",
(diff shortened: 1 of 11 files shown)


yahoocla:

mridgway:
I think we want `react-create-class` to be a `dependency` rather than a `peerDependency`. The reason being: not all users of Fluxible will use `react-create-class` since they may be using ES6 classes. Also, I don't think we'll have to worry about breaking changes in the `react-create-class` since Facebook has recommended use of ES6 classes moving forward.

J-Dickson:
@mridgway I agree with the `peerDependency`. Would you be able to clarify the second point you made please?

mridgway:
Sorry I should have clarified that one. It was another argument for using `dependencies`: one of the reasons we use `peerDependencies` is to make sure that the user is on the right version of `react`, `react-dom`, or `fluxible`. I don't think we have to worry about there being incompatibility between `react-create-class` and  React 16 for instance, so it should be ok to use `dependencies`.

J-Dickson:
@mridgway I have moved the create-react-class to `dependencies` and removed it from `peerDependencies`

redonkulus:
Thanks @J-Dickson. I tested the changes locally and they look good.

PR #531, opened 2017-04-17
Suppress deprecation message 

geta6:
Replace `React.PropTypes` with `prop-types` package to suppress following deprecation message.

```
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
```

### DIFF
--- a/packages/fluxible-addons-react/package.json
+++ b/packages/fluxible-addons-react/package.json
@@ -19,11 +19,13 @@
   },
   "peerDependencies": {
     "fluxible": ">=1.0.0",
+    "prop-types": "^15.5.8",
     "react": "^0.14.0 || ^15.0.0-rc.1",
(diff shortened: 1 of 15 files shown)


packages/fluxible-addons-react/package.json
```diff
@@ -19,11 +19,13 @@
   },
   "peerDependencies": {
     "fluxible": ">=1.0.0",
+    "prop-types": "^15.5.8",
```
mridgway: I think this should be `dependencies` similarly to what was done with `create-react-class`.

geta6: ok

redonkulus:
Thanks!

redonkulus:
@geta6 fluxible-addons-react@0.2.11 and fluxible-router@1.2.0 are now published. Thanks again!

PR #536, opened 2017-05-09
Use errorHandler for exceptions thrown inside stores

ash14:
I encountered a case in my app where certain exceptions were not handled by anything and bubbled up all the way to the main process.

I'm not sure what the intention was behind [re-throwing the error](https://github.com/yahoo/fluxible/blob/3b31d1f/packages/dispatchr/lib/DispatcherContext.js#L113), but this PR allows the `errorHandler` callback, if supplied, to be used in these cases as well.


redonkulus:
published in dispatchr@1.2.0, thanks!

PR #537, opened 2017-05-15
Remove second setImmediate wrapper around callback during action exec…

lingyan:
…ution, for less artifial yielding for most of the normal cases where callback function executes successfully with no exception.

Based on https://github.com/yahoo/fluxible/pull/518 and the differences are:

* This is a less ambitious version of https://github.com/yahoo/fluxible/pull/518, which we had some regression while testing an internal app.  The goal is the same, removing one setImmediate for normal (no-error) action execution.
* The optimization is behind a config `optimizePromiseCallback` (of `FluxibleContext` constructor) to ensure we don't accidentally break other apps using Fluxible.

@redonkulus @mridgway @kfay @ericf

### DIFF
--- a/packages/fluxible/lib/FluxibleContext.js
+++ b/packages/fluxible/lib/FluxibleContext.js
@@ -16,9 +16,17 @@ require('setimmediate');
  * A request or browser-session context
  * @class FluxibleContext
  * @param {Fluxible} app The Fluxible instance used to create the context
+ * @param {Object} options The options sharable by the context and context plugins
+ * @param {Boolean} options.optimizePromiseCallback Whether to optimize Promise callback.
+ *          Defaults to `false`. `FluxibleContext` uses two setImmediate in utils/callAction
+ *          when executing every action.  The second `setImmediate` wraps callback execution
+ *          to make sure exceptions thrown during callback execution are not swallowed by Promise.
+ *          This optimization eliminates the second `setImmediate` by catching errors caught by
+ *          Promise and throwing it. This way, successful callback executions won't need this extra
+ *          yielding because of the `setImmediate`.
  * @constructor
  */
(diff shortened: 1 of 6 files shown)


packages/fluxible/lib/FluxibleContext.js
```diff
@@ -311,6 +322,9 @@ FluxContext.prototype.dehydrate = function dehydrate() {
     var self = this;
     var state = {
         dispatcher: (this._dispatcher && this._dispatcher.dehydrate()) || {},
+        options: {
+            optimizePromiseCallback: this._optimizePromiseCallback
+        },
```
lingyan: dehydrate the option, so that client side action execution is also controlled by this config.  And since this flag can be used by any plugin or context, it is good to be dehydrated as part of context itself (versus adding a fluxible plugin for this and dehydrating there)

packages/fluxible/utils/callAction.js
```diff
@@ -36,14 +37,7 @@ function callAction (actionContext, action, payload, done) {
...
     if (done) {
-        executeActionPromise
-            .then(function(result) {
-                // Ensures that errors in callback are not swallowed by promise
-                setImmediate(done, null, result);
-            }, function (err) {
-                // Ensures that errors in callback are not swallowed by promise
-                setImmediate(done, err);
-            });
+        promiseCallback(executeActionPromise, done, {optimize: actionContext.optimizePromiseCallback});
```
lingyan: Extracting this to a util function so that other places (where promises need to call a callback) can use the same code.

lingyan:
Build failed because latest jsdom does not work with node4.  I've updated master to remove node4 from build.  Given all tested passed with node 6 and node 7, I am merging this PR.

PR #542, opened 2017-05-23
Switch to the prop-types package

kfay:
@Vijar @redonkulus @lingyan 

This removes a react 15.5 warning about `PropTypes` by switching to the `prop-types` package rather than using `React.PropTypes`


redonkulus:
Thanks!

PR #550, opened 2017-06-28
Minor fixes

tibbes:
I found these opportunities for minor fixes while testing [lgtm](https://lgtm.com/projects/g/yahoo/fluxible/alerts/). I hope this is useful.

### DIFF
--- a/packages/fluxible-router/lib/handleHistory.js
+++ b/packages/fluxible-router/lib/handleHistory.js
@@ -41,9 +41,6 @@ if (HAS_PUSH_STATE) {
     window.addEventListener(EVENT_POPSTATE, preloadListener);
 }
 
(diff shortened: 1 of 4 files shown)


packages/fluxible-router/lib/handleHistory.js
```diff
@@ -77,11 +74,6 @@ function createComponent(Component, opts) {
     Object.assign(HistoryHandler.prototype, {
 
         componentDidMount: function () {
-            if (historyCreated) {
-                throw new Error('Only one history handler should be on the ' +
-                'page at a time.');
-            }
-
```
lingyan: haha. guess we never set `historyCreated` to `true` as the logic intended.  i'm fine with removing it :)

PR #563, opened 2017-09-28
Move files to src folder, compile into previous location for back compat.

snyamathi:
@redonkulus @lingyan If there would be changes to `fluxible-router` that require transpilation to maintain backward compatibility, then we'll need to move the files to a src folder.

If we were to transpile into a `dist` folder, then we'd need to break the API and do a major version bump, which would be a pain.  We could, instead, put the transpiled files back where they would be - but this is somewhat of a departure from other packages like `fluxible-plugin-devtools` which uses a `dist` folder.  However, this is basically what `react-intl` does - it transpiles from the `src` folder into the package root.

Next time we need to do a major version bump to this pacakge, we should make that folder structure change, but this should be fine for now.

Note that I had to remove the `rm -rf` command run on the dist location, since that would wipe out our package as well as modify the gulp command do accept a param `-d` for where in the package we should put the dist files.

I chose `-d` to match the babel-cli https://babeljs.io/docs/usage/cli/#babel-compile-directories but open to changing that.

This sets the groundwork required to migrate this packages components from `create-react-class` to an es6 class.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@ dist
 artifacts
 node_modules
 *.log
+*.tgz
 .jshintrc
 package-lock.json
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -61,9 +61,9 @@ gulp.task('dist', () => {
     if (!packageName) {
         throw new Error('Usage: gulp dist -p <package>');
     }
+    let outDir = argv['out-dir'] || argv.d || '';
     var src = packages[packageName] + '/src/**/*';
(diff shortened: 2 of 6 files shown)


redonkulus:
@snyamathi since we may bump fluxible-router major version due to https://github.com/yahoo/fluxible/pull/565, we could make these changes as well. Internally, the changes shouldn't be too bad since we have one central library that pulls in the fluxible-router ones.

This branch would need to be rebase since the gulp stuff is gone now.

snyamathi:
Closed in favor of https://github.com/yahoo/fluxible/pull/585

Issue #546, opened 2017-06-20
Replace createReactClass with ES6 classes


alexparish:
The module _create-react-class_ is a dependency of both _fluxible-addons-react_ and _fluxible-router_ and as such will bloat any projects that consume these packages.

Instances of createReactClass should be replaced with ES6 classes that extend React.Component such that the dependency on the _create-react-class_ module can be removed.

redonkulus:
@alexparish Good point.  For fluxible-addons-react that package is actually only used for testing, so we could move the dependency to devDependencies to avoid the extra bloat.

For fluxible-router, we could refactor that away too. We would have to ensure we are not leveraging mixins or anything to ensure we do not break functionality.

redonkulus:
Published in fluxible-addons-react@0.2.13

alexparish:
This issue seems more pressing now React 16 has launched. 

This latest version of React has removed the dependency on _create-react-class_ but `fluxible-router` still has a dependency on it.

redonkulus:
@alexparish Shouldn't be an issue with React 16 because we are using the [create-react-class package](https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/package.json#L28) instead of relying on React.createReactClass directly.

alexparish:
@redonkulus Sure, I understood that nothing will break. 

My concern was the bloat of bringing in that extra package. Removing the dependency (and using ES6 classes instead) would shave 12.5K from my bundle. Not huge, but it seems an unnecessary dependency. 

I will look into creating a PR for this if I can find time.

mridgway:
I believe there is only one place that uses `createReactClass` at this point: https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/lib/createNavLinkComponent.js

I migrated everything to use ES5 classes in https://github.com/yahoo/fluxible/commit/9c46478e2fb9cb11f942a608db969af2e425abcb but did not migrate that one case for a reason I can't remember. It may be a backward incompatibility but I can't remember.

alexparish:
I have created a PR to remove that final instance of `createReactClass`

alexparish:
Updated PR created: https://github.com/yahoo/fluxible/pull/586

PR #565, opened 2017-09-28
Remove create-react-class dependency

alexparish:
See issue https://github.com/yahoo/fluxible/issues/546

### DIFF
--- a/packages/fluxible-router/lib/createNavLinkComponent.js
+++ b/packages/fluxible-router/lib/createNavLinkComponent.js
@@ -10,7 +10,6 @@ var RouteStore = require('./RouteStore');
 var debug = require('debug')('NavLink');
 var navigateAction = require('./navigateAction');
 var __DEV__ = process.env.NODE_ENV !== 'production';
-var createReactClass = require('create-react-class');
 
 function objectWithoutProperties(obj, keys) {
     var target = {};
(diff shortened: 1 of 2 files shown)


redonkulus:
Thanks @alexparish! We will need to test this internally before we merge. We will probably have to bump fluxible-routers major version because internally we do some overrides that do not assume the class structure.

alexparish:
Sure no worries and no rush!

Also I have little knowledge of the component I worked on so if the unit tests aren't comprehensive then there's a good chance I may have inadvertently broken a use case. So definitely worth treading carefully with this one!

alexparish:
Hi @redonkulus. Any progress on your internal testing?

redonkulus:
> Hi @redonkulus. Any progress on your internal testing?

Sorry not yet. The code we have internally is used in a lot of places and would need a lot of regression testing. We just haven't had the time to look into it yet. Hopefully sooner rather than later.

alexparish:
I don't suppose you've looked at this any further? I'm not using fluxible anymore but I think it would beneficial people who are.

redonkulus:
Sorry I dropped the ball on this one, I will see if I can find some time to take a look. I may just bump the major version to avoid any uncaught issues.

Can you rebase to fix the conflicts?

alexparish:
@redonkulus Sure, major version may be safer.

Rebase was messy so I've recreated the PR from the current master branch: https://github.com/yahoo/fluxible/pull/586

PR #566, opened 2017-10-01
Support React 16

redonkulus:
@lingyan @snyamathi @mridgway 

See https://reactjs.org/blog/2017/09/26/react-v16.0.html

Upgrades:
- fluxible-addons-react
- fluxible-router
- generator-fluxible
- site
- tests


site/tests/unit/actions/demoException.js
```diff
@@ -18,10 +18,9 @@ describe('site', () => {
 
         it('should execute the action', function (done) {
             demoException(context, {}, function (err) {
-                expect(err).to.eql(new Error('Whoops!'));
+                expect(err).to.eql(new Error('WWhoops!'));
```
mridgway: Is this legit?

redonkulus: Doh that was a test ;) fixing...

alexparish:
Great to see!

Could this and https://github.com/yahoo/fluxible/pull/565 be rolled into major releases for all packages?

redonkulus:
@alexparish I will keep them separate as the fluxible router change requires more scrutiny and shouldn't block this PR from being released. 

Plus, I do not plan to bump major for this PR as there should not be any breaking changes. Bumping major is done more carefully as that requires us to update a lot of packages internally.

redonkulus:
Released in fluxible-router@1.4.0 and fluxible-addons-react@0.2.14

PR #567, opened 2017-10-02
Add lerna

redonkulus:
@mridgway @lingyan @snyamathi 

This deprecates the custom gulp setup with lerna. We have used lerna successfully internally for a few monorepo's and it makes things like installing, linking cross dependencies and testing easier. The version/publish subtasks from the gulp file can also be replaced by lerna.

One thing we will miss is the per package tagging that we use today. This is a nice to have, but not a blocker for migration. We will just use a global tag for all changes (important because the `--since` flag uses git tags to determine package changes). Unfortunately, lerna doesn't support tags per package.

The commits describe the changes done.


packages/fluxible-plugin-devtools/package.json
```diff
@@ -9,11 +9,11 @@
   },
   "scripts": {
     "pretest": "npm run dist",
-    "test": "../../node_modules/.bin/mocha tests/unit/ --recursive --compilers js:babel-register --reporter spec",
+    "test": "../../node_modules/.bin/mocha tests/unit/ --opts ../../mocha.opts",
```
lingyan: 👍

package.json
```diff
@@ -27,30 +29,36 @@
...
-    "coveralls": "^2.11.1",
+    "coveralls": "^3.0.0",
+    "create-react-class": "^15.5.1",
     "es6-promise": "^4.0.2",
     "eslint": "^4.1.0",
     "eslint-plugin-babel": "^4.0.0",
     "eslint-plugin-react": "^7.0.0",
+    "fluxible": "^1.0.0",
     "gulp": "^3.9.0",
     "gulp-babel": "^7.0.0",
```
lingyan: still need gulp deps?

snyamathi:
+1 on Lingyan's comment.  My only concern is that by using `babel-register` as an option to mocha for all unit tests that it's possible for code which would normally be required to be transpiled will test fine, but break when used on browsers that don't understand such features.

For example, class properties are pretty widely not supported, but we're transpiling them for tests.  That's fine, so as long as when the module is consumed downstream that it's always fronted by babel or some other transpilation step - and not everyone will transpile their node_modules.

package.json
redonkulus: Nah, I'm going to remove it

redonkulus:
@snyamathi I think should be ok as we already recommend to use a transpiler on the readme when consuming these packages. I believe the fluxible-addons-devtools package uses some newer language stuff but that gets transpiled before publishing.

packages/fluxible-addons-react/FluxibleComponent.js
```diff
@@ -2,6 +2,7 @@
  * Copyright 2015, Yahoo Inc.
  * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
+/*eslint no-unused-vars:0 */
```
mridgway: Why did this fail?

mridgway:
Seems like the eslint rules were double disabled here. The base eslintrc disabled a bunch of rules that I think are also individually disabled in specific files.

packages/fluxible-addons-react/FluxibleComponent.js
redonkulus: Because of:
```
function FluxibleComponent(props, context) {
    React.Component.apply(this, arguments);
}
```

`props` and `context` are not used. I wanted to leave those in to help document the API. So I just disabled the rule for that file.

redonkulus:
> Seems like the eslint rules were double disabled here. The base eslintrc disabled a bunch of rules that I think are also individually disabled in specific files.

Ya I think this happened because I started disabling individual lines, then just did it from the root rc file. I can clean that up.

lingyan:
@redonkulus one more: update README?

redonkulus:
> @redonkulus one more: update README?

fixed @lingyan

PR #571, opened 2017-10-11
typo fix

ns3777k:


lingyan:
Nice catch!  Thanks for fixing @ns3777k  👍

PR #573, opened 2017-11-08
Remember the referrer url in history state object

Nagarkarricha:
## Reviewers
@lingyan @redonkulus 

## Goal
Remember the referrer url in the history. This honors the correct referrer for all client side navigations and particularly helpful for handling back button clicks during page navigation correctly. 

_Example:
Page:A ( origUrl: /A.html ) -> Page:B ( origUrl: /B.html, referrerUrl: https://www.example.com/A.html ) -> C.html ( origUrl: /C.html, referrerUrl: https://www.example.com/B.html )_

The history object would look something like this now:
```
history = {
    length: 3,
    scrollRestoration: manual,
    state: {
        origUrl: "abc.html",
        referrerUrl: "http://www.xyz.html",
        scroll: {
            x: 0,
            y: 0
        }
    }
}
```

### DIFF
--- a/packages/fluxible-router/lib/History.js
+++ b/packages/fluxible-router/lib/History.js
@@ -102,7 +102,9 @@ History.prototype = {
             url = isUndefined(url) ? win.location.href : url;
 
             // remember the original url in state, so that it can be used by getUrl()
-            var _state = Object.assign({origUrl: url}, state);
+            // remember the referrer url in state, so that it can be used for back navigations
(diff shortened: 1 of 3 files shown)


lingyan:
@Nagarkarricha Code looks good.  Could you bump patch version in fluxible-router package.json?

yahoocla:

yurited:
👍

PR #586, opened 2018-04-16
Remove create-react-class dependency

alexparish:
See issue #546


redonkulus:
@alexparish released in 2.0.0-beta.0 as I test.

PR #594, opened 2018-06-28
Update hoist-non-react-statics to the latest version

kfay:
@yahoo/team-fluxible @mridgway 

This updates hoist-non-react-statics to the latest version compatible with react 16.3 https://github.com/mridgway/hoist-non-react-statics#compatible-react-versions


kfay:
@mridgway I just want to confirm that there are no backward incompatibility concerns here since the major version was bumped.

mridgway:
There are not any incompatibilities AFAIK. The major version bump was for a change that has since been reverted.

PR #612, opened 2019-04-18
expose unbuilt src via the package module key

pwmckenna:
we preserve the existing fluxible-router export by moving index.js into src and updating the main field in package.json to point into dist, but this also allows us to expose the unbuilt code via the module key as suggested by webpack :https://webpack.js.org/guides/author-libraries/#final-steps


snyamathi:
I do like the idea of exporting files with ES2015 module syntax in the `modules` field of the package.json - basically what the link you have posted recommends.

However, in this particular library, everything is CommonJS and we simply use babel to transpile [a small portion of the syntax](https://github.com/yahoo/fluxible/blob/master/.babelrc).  I'm not sure that we would want to export CommonJS modules via the `module` field.

If we were to move the whole package to ES Modules, we should be careful about the nuances of esm vs commonjs.  For example here https://github.com/yahoo/fluxible/blob/master/packages/fluxible/utils/index.js#L1 we want to be careful with exporting object vs using named exports.

redonkulus:
Agree with @snyamathi here. We should only do this once all code is migrated to es modules. Otherwise, it will be confusing to users.

@pwmckenna If you want to open a new PR that converts to using import/exports then I think this change would make more sense.

Issue #615, opened 2019-05-29
Should dispatch NAVIGATE_FAILURE if action is not able to resolve


lingyan:
https://github.com/yahoo/fluxible/blob/7ad54c83ee9694af8733918b32fc38bdec057ecf/packages/fluxible-router/src/lib/navigateAction.js#L49-L54

There are two cases here:
* `action` is not specified (`route.action` is empty): Fine to dispatch `NAVIGATE_SUCCESS`
* `action` is specified but unable to resolve into a function: **SHOULD dispatch `NAVIGATE_FAILURE`**

We should fix the second case to dispatch failure event.  Dispatching success event masks the problem and makes debugging hard.

redonkulus:
This was fixed in both fluxible-router@1.x. and 2.x major versions.

lingyan:
Yeah! thanks @redonkulus 🥇

PR #616, opened 2019-06-18
[fluxible-router] Return failure when no action can be resolved

tom76kimo:
@lingyan @redonkulus 

Based on #615 

Return failure when no action can be resolved


packages/fluxible-router/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fluxible-router",
-  "version": "2.0.0-beta.3.0",
+  "version": "2.0.0-beta.4.0",
```
redonkulus: FYI this change will not make it into 1.x branch, which we use internally. If you need it for existing apps, you may need to backport this change to 1.x.

packages/fluxible-router/src/lib/navigateAction.js
```diff
@@ -42,13 +42,24 @@ function navigateAction (context, payload, done) {
...
     if (!action || 'function' !== typeof action) {
-        debug('route has no action, dispatching without calling action');
-        context.dispatch('NAVIGATE_SUCCESS', completionPayload);
+        debug('action cannot be resolved');
+        completionPayload.error = {
+            statusCode: 500,
+            message: 'Action for ' + payload.url + 'can not be resolved'
+        };
+        context.dispatch('NAVIGATE_FAILURE', completionPayload);
         done();
```
lingyan: we should change `done()` to pass the error

tom76kimo:
@lingyan @redonkulus , updated! Could you help take a look again? Thank you.

packages/fluxible-router/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fluxible-router",
-  "version": "2.0.0-beta.4.0",
+  "version": "1.5.4",
```
redonkulus: @tom76kimo  This won't work as the master branch has breaking changes for 2.x. You would need to branch from the 1.x branch and push your changes there if you want it released in the 1.x.x branch.

tom76kimo: Ah, you are right. Will fix it.

pablopalacios:
I just got in the issue as well. @tom76kimo if you don't mind, I can rebase your work on top of 1.X and move this issue forward.

tom76kimo:
@pablopalacios Please do, thank you!

pablopalacios:
@lingyan @redonkulus it looks like we don't have a 1.x.x branch anymore, only the tag. I already rebased @tom76kimo patch on top of that tag but I cannot create a PR for it since there would be no target branch. So, would you mind to recreate the branch so we can have an 1.5.4 version released. I can also backport the change to master if it is needed but a 1.5.4 version would be awesome as well.

lingyan:
@pablopalacios Thanks for following up on this fix!

The code change in this PR looks good, except the version update: https://github.com/yahoo/fluxible/pull/616/files#r393192433

Once that is fixed. we can merge this PR. and was talking with @redonkulus, he can create a 1.x branch for porting the fix to the 1.x version.

redonkulus:
@pablopalacios this has been back ported and published under fluxible-router@1.5.4.

PR #623, opened 2019-11-08
[fluxible-router] Catch onbeforeunload access errors

dplate:
In Firefox it could be forbidden to call onbeforeunload if it wasn't set by the page itself, e.g. set by plugins. When calling it, it will throw an Error with the message "Permission denied to access object". The result is, that client side navigation will not work anymore.
Since I couldn't find a possibility to check if the function is allowed to be called, I catch any execption and only log the error.

An example plugin which causes this issue:
https://github.com/tjhrulz/WebNowPlaying

### DIFF
--- a/packages/fluxible-router/lib/createNavLinkComponent.js
+++ b/packages/fluxible-router/lib/createNavLinkComponent.js
@@ -277,7 +277,15 @@ module.exports = function createNavLinkComponent (overwriteSpec) {
 
             e.preventDefault();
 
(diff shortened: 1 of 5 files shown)


lingyan:
@dplate do you mind porting this fix to the master branch as well?  thanks!

dplate:
Thanks for integrating it.
Here the port for the master branch:
https://github.com/yahoo/fluxible/pull/624

Issue #636, opened 2020-05-20
Update React lifecycle methods


BryceV:
In a couple components there are old react lifecycle methods. In my case it was componentWillReceiveProps in connectToStores:

<img width="523" alt="Screen Shot 2020-05-20 at 8 56 21 AM" src="https://user-images.githubusercontent.com/9063864/82468805-18243300-9a78-11ea-8344-d865745646e1.png">

pablopalacios:
@redonkulus this is a duplicate of #587 and I'm already solving it. Feel free to close it.

redonkulus:
Sounds good to me.

Issue #587, opened 2018-04-25
componentWillReceiveProps to getDerivedStateFromProps


roderickhsiao:
React 16.3 introduce `getDerivedStateFromProps`. currently we are using in `connectToStores.js` and `createNavLinkComponent.js`.

React already starts throwing warning if client migrate the component to use `getDerivedStateFromProps` while wrapping with `connectToStores` 

![screen shot 2018-04-25 at 6 14 47 am](https://user-images.githubusercontent.com/3906130/39248072-fee78990-484f-11e8-8689-40fd943a43c1.png)

We could use https://github.com/reactjs/react-lifecycles-compat#readme to support old version.

redonkulus:
Using the polyfill is probably the safest bet so that we don't have to bump major version. I will try to get this when I get time or you can provide a PR 😀

roderickhsiao:
the honor is yours :) I will check if I have time as well, I think polyfill should be fine but we probably want to wait the polyfill to be stable as it do have breaking changes for the past two weeks

https://github.com/reactjs/react-lifecycles-compat/blob/master/CHANGELOG.md

redonkulus:
@pablopalacios looks like in the new 1.x release we still have the `UNSAFE_componentWillReceiveProps` prefix in the places mentioned above, from your [analysis](https://github.com/yahoo/fluxible/pull/650) in this change, looks like we cannot use `getDerivedStateFromProps` since its a static and we need access to `this.context` (although maybe this has changed since last year). 

What are your thoughts now that we have a 1.x release and could remove the old UNSAFE_ methods.

pablopalacios:
It might be possible if we convert `connectToStores` to a functional component and use the `useEffect` hook. But I bet that it's still a tricky problem. I might give a try soon and see if it's possible.

redonkulus:
Fixed via https://github.com/yahoo/fluxible/pull/747

Issue #584, opened 2018-04-06
[addons-react] Following React 16.3+ new Context API ?


fsubal:
### Expected behavior
React 16.3+ has now [new Context API](https://reactjs.org/docs/context.html) and it is deprecating legacy one like `this.context = ...` or `static contextTypes`. Maybe functions like `executeAction` and `getStore` should be provided with the new way.

### Actual behavior
fluxible-addons-react does not provide those based on new context API. It still depends on the legacy one.

### Information about the Issue

https://reactjs.org/docs/context.html

### Steps to reproduce the behavior
None.

redonkulus:
@fsubal We will have to take a look at the new API and see how that changes the implementation. I believe React is keeping the existing `context` API for sometime, but that might be just until v17 is released.

pwmckenna:
I don't believe it will be possible to upgrade to the new context api without a breaking change. We'll need some globally shared context object (`react.createContext()`) that both fluxible-addons-react can use for connectToStores etc, but that object must also be assessible to your app when injecting the context. If you could guarantee that everyone was already using exclusively `provideContext` to inject the fluxible context this would be possible, but we don't really have that guarantee.

I think that one of the large drivers of this upgrade will be hooks. `useContext` uses exclusively the new api, so if you want to use `useContext` you'll have to define your own new context object, and use the provider to inject that *in addition* to whatever you're doing currently to get context down to your `connectToStores` etc. That doesn't like a terrible interm solution in my mind.

The one thing I believe we should do immediately to help future proof fluxible-addons-react would be to actually expose *the* new fluxible context (along with some docs on how to inject it along side your existing context), even if its not using it internally. This would allow fluxible based libraries to upgrade to the latest context api, without introducing a doomed from the start context object that will never be used by `connectToStores`. It would make the future upgrade path much smoother, allowing you to opt into the double fluxible injection at your leisure, then once fluxible-addons-react was upgraded, there would be nothing to do even if it was a "breaking" change.

The PR will be tiny, so I'll go ahead and just create it, regardless of whether folks think this is a good idea.

pwmckenna:
[...]

PR #646, opened 2020-09-05
Refactored fluxible-addons-react to use ES2015+ features

pablopalacios:
So, before tackling #636, #587, #584 and introduce any breaking changes, I thought it would be good to make fluxible react components to look like modern react components. So I basically just converted all components to be ES2015 classes.

### DIFF
--- a/packages/fluxible-addons-react/package.json
+++ b/packages/fluxible-addons-react/package.json
@@ -18,7 +18,6 @@
   },
   "dependencies": {
     "hoist-non-react-statics": "^3.3.2",
-    "inherits": "^2.0.4",
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -51,10 +51,6 @@
     "sinon": "^9.0.1",
     "yargs": "^16.0.0"
   },
-  "devEngines": {
-    "node": "4.x || 5.x || 6.x",
-    "npm": "2.x || 3.x"
-  },
   "pre-commit": [
     "dev:lint",
     "dev:test"
--- a/packages/fluxible-addons-react/src/FluxibleComponent.js
+++ b/packages/fluxible-addons-react/src/FluxibleComponent.js
@@ -4,42 +4,32 @@
  */
 'use strict';
 
-var React = require('react');
-var PropTypes = require('prop-types');
-var inherits = require('inherits');
+const React = require('react');
+const PropTypes = require('prop-types');
 
-function FluxibleComponent(props, context) {
-    React.Component.apply(this, arguments);
-}
-
-inherits(FluxibleComponent, React.Component);
-
-FluxibleComponent.displayName = 'FluxibleComponent';
-FluxibleComponent.propTypes = {
-    context: PropTypes.object.isRequired
-};
-FluxibleComponent.childContextTypes = {
-    executeAction: PropTypes.func.isRequired,
-    getStore: PropTypes.func.isRequired
-};
-
-Object.assign(FluxibleComponent.prototype, {
-    /**
-     * Provides the current context as a child context
-     * @method getChildContext
-     */
-    getChildContext: function () {
+class FluxibleComponent extends React.Component {
+    getChildContext() {
         return {
             getStore: this.props.context.getStore,
             executeAction: this.props.context.executeAction
         };
-    },
+    }
 
-    render: function () {
+    render() {
         return React.cloneElement(this.props.children, {
             context: this.props.context
         });
     }
-});
+}
+
+FluxibleComponent.propTypes = {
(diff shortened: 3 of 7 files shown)


pablopalacios:
@redonkulus I also removed "devEngines" from the root package.json given that it was completely outdated according to the travis configuration.

redonkulus:
Have you tested this locally with one app of yours to see if everything works properly?

pablopalacios:
> Have you tested this locally with one app of yours to see if everything works properly?

Yup! I did 2 tests: first I generated a fluxible app with yo but I also tested with my company's big app (we have more than 50 stores there) and I had no issues.

redonkulus (approved):
LGTM

packages/fluxible-addons-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fluxible-addons-react",
-  "version": "0.2.16",
+  "version": "0.2.17",
```
lingyan, line 3: we should bump this to 1.0.0, since using `class` means all apps using fluxible-addons-react need to transpile to ES5 for old browsers.  I hope they do already. But would be good to be on the safe side to bump major here.

packages/fluxible-addons-react/package.json
```diff
@@ -14,13 +14,12 @@
...
   "dependencies": {
     "hoist-non-react-statics": "^3.0.1",
-    "inherits": "^2.0.1",
     "prop-types": "^15.5.8"
   },
   "peerDependencies": {
     "fluxible": ">=1.0.0",
     "react": "^0.14.0 || ^15.0.0 || ^16.0.0",
-    "react-dom": "^0.14.0 || ^15.0.0 ||  || ^16.0.0"
+    "react-dom": "^0.14.0 || ^15.0.0 || ^16.0.0"
```
lingyan, line 22: if we are bumping major version, we can remove `^0.14.0` from this line and the line above

lingyan (commented):
LGTM.  two minor comments.

packages/fluxible-addons-react/package.json
redonkulus: @pablopalacios we could even make this a 1.0.0-beta.0 version and add in more breaking changes that we need. if you are ok with that, you can remove the version bump from this PR.

pablopalacios:
Thanks @redonkulus and @lingyan for the review :)

So, after thinking more about this PR and taking in consideration your comments I would do the following:

1. Merge #647 first (no breaking changes, just tooling update)
2. Create another PR adding the react lifecycle polyfill to all required packages and release as minor change (solves #636 and #587 , doesn't introduce breaking changes and allows people to run fluxible with older and newer versions of react 17).
3. Create another PR that would set babel for addons-react and release as minor (no break changes, just setting up tooling in the same way we do with fluxible-plugin-devtools)
4. Merge this PR and release as minor (again, no breaking changes, just modernizing but keeping compatibility with old browsers)
5. We could do this for all the packages we want and keep releasing as minor.

Then I would say that we could move to a new major version and do the following breaking changes:

- Remove polyfill and drop old react support
- Remove decorator support
- Remove mixins
- Update to use the newer context API (solves #584 )

What do you think about this plan? It would not be an issue for me to prepare all these PRs.

redonkulus:
I think that looks like a good plan. The biggest thing for us is to not introduce any breaking changes since whatever we publish here, will get pulled in by our internal apps. We appreciate you taking the time to help modernize the code base and move things along.

Also, I'm assuming when you said "release minor", that you meant patch version for 0.x.x package versions?

pablopalacios:
@redonkulus exactly. I'm just following semver here. But I also mean no beta releases until we conclude the 5th step. The point of these minor/patch releases would beto  just make fluxible compatible with React 17 without breaking the API but also keeping compatibility with old browsers/react versions.

redonkulus:
@pablopalacios want to rebase this against the new fluxible-addons-react-1.x.dev branch?

pablopalacios:
@redonkulus yes, I will. But I would like to add babel support first so we can play safe (and also add jsx support). Would it be fine with you?

redonkulus:
That’s fine by me. I thought we already integrated Babel.

redonkulus (approved):
LGTM

redonkulus:
Should we merge this in now or do you want to make more changes?

pablopalacios:
We are still not using babel here. There will be another PR before we can merge this one. As soon as it's done, I'll remove the wip label and ask you again for review ;)

pablopalacios:
@redonkulus @lingyan would you have any further comments on this one? I'm finished with my changes here ;)

redonkulus (approved):
Great job!

PR #647, opened 2020-09-06
Updated babel to latest version (and fixed coveralls integration as a bonus)

pablopalacios:
Summary of changes:
- updated babel to version 7
- centralized babel configuration in one place (I can revert that, no worries, but I thought it would be easier to maintain if we have just one configuration)
- fixed integration between nyc, travis and coveralls (actually, coveralls was complaining about babel version, this is why I came up with this PR)
- uncommented fluxible-addons-react tests that were not working with babel 6


pablopalacios:
By the way, @redonkulus , is there any reason why we don't use package-lock?

redonkulus:
@pablopalacios with monorepo the packages installed can change based the packages that were changed. so that might mess up the package lock across different builds.

packages/dispatchr/package.json
```diff
@@ -8,9 +8,9 @@
     "url": "git://github.com/yahoo/fluxible.git"
   },
   "scripts": {
-    "test": "../../node_modules/.bin/mocha tests/unit/",
-    "cover": "BABEL_ENV=test ../../node_modules/.bin/nyc ../../node_modules/.bin/_mocha tests/unit/",
-    "lint": "../../node_modules/.bin/eslint lib/ addons/ utils/ index.js"
+    "test": "mocha tests/unit/",
```
redonkulus, line 11: we've always used the .. notation because if you try to run the command in the package folder it fails since npm can't find the binary locally.

packages/fluxible-addons-react/tests/unit/lib/connectToStores.js
```diff
@@ -91,62 +91,58 @@ describe('fluxible-addons-react', function () {
...
-            //
-            //expect(appContext.getStore(BarStore).listeners('change').length).to.equal(1);
-            //expect(appContext.getStore(FooStore).listeners('change').length).to.equal(1);
-            //
-            //ReactDOM.unmountComponentAtNode(container);
-            //
-            //expect(appContext.getStore(BarStore).listeners('change').length).to.equal(0);
-            //expect(appContext.getStore(FooStore).listeners('change').length).to.equal(0);
+        it('should get the state from the stores using decorator pattern', function (done) {
+            @connectToStores([FooStore, BarStore], (context) => {
```
redonkulus, line 95: we should really remove support for decorators since the api is legacy and will never make it into the spec now.

redonkulus (commented):
Changes look good in general, not sure if we need to change the run scripts in each package.json. Are you just executing the commands from the repository root?

packages/fluxible-addons-react/tests/unit/lib/connectToStores.js
pablopalacios: I'm glad you said that. But that would be another breaking change, right? So, I would keep it for now since this PR just updates tooling. Then, when we move to the beta versions where we really start introducing breaking changes, I would definitely removed it.

packages/dispatchr/package.json
pablopalacios: > Changes look good in general, not sure if we need to change the run scripts in each package.json. Are you just executing the commands from the repository root?

You're completely right about not being able to run the commands inside each package with this patch.

I didn't realize it because I was always running `lerna test` (which I do most of the time since the tests are fast and one package can break another) or `lerna test --scope package-name` (to run all the tests for a given package) or even `npx mocha packages/package-name/tests/test-file.js` (for a specific test file) always from the monorepo root.

So we have 2 alternatives:

1. Keep as it was before: polluted, but allows running tests from inside each package
2. Keep as it is: cleaner, but doesn't allow running tests inside each package (lerna, mocha and eslint allow us to specify target packages/files anyway)

I'm ok with both (we have a "clean" X "correct" trade off), but which one would you prefer?

redonkulus: @pablopalacios lets go with how it was before. The syntax we had doesn't break how lerna is used at the root. It's a nice convenience for us to be able to run tests from within the package directory.

pablopalacios: Reverted.

redonkulus (approved):
awesome work!

PR #649, opened 2020-09-12
fluxible: remove react unsafe methods

pablopalacios:
Removing them from `fluxible-react-addons` is a little bit tricky since `getDerivedStateFromProps` is a static method so we wouldn't have access to the context. But `fluxible` package also had a component with the deprecated method that was easy to replace. So I'm delivering it now because I will need more time in the other one.


redonkulus (approved):
LGTM

PR #650, opened 2020-09-12
fluxible-addons-react: remove depreacted react methods

pablopalacios:
This is related to #587 and #636. Here I'm just solving the issue for fluxible-addons-react.

This one was a tricky. We can't use `getDerivedStateFromProps` since it is static and we need to have access to `this.context`. 

The solution that I found was to use `componentDidUpdate` since react assures that user will not see the "in between" state. But since `setState` triggers `componentDidMount`, we needed a condition to prevent an infinite recursion during an update. 

So I added a key in the state so we have something as a reference if the component was already updated or not. This key is invalidated whenever an update occurs. The key is random generated by nanoid. I opted for nanoid since it is even smaller than the react polyfill we were planning to use.

I haven't worked on fluxible-router yet since I would like to know first if this approach is valid or not.

### DIFF
--- a/packages/fluxible-addons-react/connectToStores.js
+++ b/packages/fluxible-addons-react/connectToStores.js
@@ -8,6 +8,7 @@ var React = require('react');
 var PropTypes = require('prop-types');
 var inherits = require('inherits');
 var hoistNonReactStatics = require('hoist-non-react-statics');
+var nanoid = require('nanoid').nanoid;
 
 function createComponent(Component, stores, getStateFromStores, customContextTypes) {
     var componentName = Component.displayName || Component.name;
@@ -41,12 +42,16 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
                 this.context.getStore(Store).removeListener('change', this._onStoreChange);
             }, this);
         },
-        componentWillReceiveProps: function componentWillReceiveProps(nextProps){
-            this.setState(this.getStateFromStores(nextProps));
+        componentDidUpdate: function componentDidUpdate(prevProps, prevState) {
+            if (this.state._key === prevState._key) {
+                this.setState(this.getStateFromStores());
+            }
         },
         getStateFromStores: function (props) {
-            props = props || this.props;
-            return getStateFromStores(this.context, props);
+            return {
+                fromStore: getStateFromStores(this.context, this.props),
+                _key: nanoid()
+            };
         },
         _onStoreChange: function onStoreChange() {
             if (this._isMounted) {
@@ -55,7 +60,7 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
         },
         render: function render() {
             var props = Component.prototype && Component.prototype.isReactComponent ? {ref: 'wrappedElement'} : null;
-            return React.createElement(Component, Object.assign({}, this.props, this.state, props));
+            return React.createElement(Component, Object.assign({}, this.props, this.state.fromStore, props));
         }
     });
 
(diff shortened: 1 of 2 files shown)


packages/fluxible-addons-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fluxible-addons-react",
-  "version": "0.2.16",
+  "version": "0.2.17",
```
redonkulus, line 3: you dont have to update version, we can do it when we publish

packages/fluxible-addons-react/connectToStores.js
```diff
@@ -41,12 +42,16 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
                 this.context.getStore(Store).removeListener('change', this._onStoreChange);
             }, this);
         },
-        componentWillReceiveProps: function componentWillReceiveProps(nextProps){
-            this.setState(this.getStateFromStores(nextProps));
+        componentDidUpdate: function componentDidUpdate(prevProps, prevState) {
+            if (this.state._key === prevState._key) {
```
redonkulus, line 46: what happens if we do not check the key?

packages/fluxible-addons-react/connectToStores.js
```diff
@@ -41,12 +42,16 @@ function createComponent(Component, stores, getStateFromStores, customContextTyp
                 this.context.getStore(Store).removeListener('change', this._onStoreChange);
             }, this);
         },
-        componentWillReceiveProps: function componentWillReceiveProps(nextProps){
-            this.setState(this.getStateFromStores(nextProps));
+        componentDidUpdate: function componentDidUpdate(prevProps, prevState) {
```
redonkulus, line 45: what kind of subtle changes do we expect to see moving from `componentWillReceiveProps` to `componentDidUpdate`?

packages/fluxible-addons-react/connectToStores.js
pablopalacios: Than we would end in a recursive loop of setState -> componentDidUpdate -> setState. From [React docs](https://reactjs.org/docs/react-component.html#componentdidupdate) 

> You may call setState() immediately in componentDidUpdate() but note that it must be wrapped in a condition like in the example above, or you’ll cause an infinite loop.

packages/fluxible-addons-react/connectToStores.js
pablopalacios: I would say that we would have mainly a performance decrease. Since updating the state after the component is updated we will have a second call to `render` method. Also from the same React documentation section posted above:

> It would also cause an extra re-rendering which, while not visible to the user, can affect the component performance.

pablopalacios: But I think we can have the same issue that react-redux has: [stale-props](https://react-redux.js.org/api/hooks#stale-props-and-zombie-children).

pablopalacios:
@redonkulus hold this one a little bit more. I've just realized that calling render twice may be an issue for a child connected component that relies in the parent props.

redonkulus:
I think in general, this change could be problematic for our internal applications. We have a lot of code that relies on these old lifecycle methods and a change like this could introduce subtle differences/issues that are hard to find. 

I would suggest we start a "1.x" branch from master with all the changes for react 16/17 and then release that in a major version (1.x) release. This will insulate us from any unwanted breakages and allow you and the rest of the community to move forward faster.

redonkulus:
I've pushed [fluxible-addons-react-1.x-dev](https://github.com/yahoo/fluxible/tree/fluxible-addons-react-1.x-dev) to the remote so that you can open PRs against this branch.

pablopalacios:
I think you are right. However we will still have the same issue if we don't use the unsafe methods. But at least we can use the UNSAFE prefix for a while and drop support for old react versions in a major update. It will be easier and safer in the end.

PR #655, opened 2020-09-26
fluxible-addons-react: integrated with babel

pablopalacios:
This one will enable us to migrate to es2015 an beyond. Then we can move #646 forward.


redonkulus (approved):
Awesome!

PR #656, opened 2020-09-27
fluxible-router:  renamed unsafe methods

pablopalacios:


pablopalacios:
@lingyan @redonkulus  Would this change be fine given that fluxible-router is already in a beta version?

lingyan (approved):
Should be fine. Thanks!

PR #663, opened 2020-10-25
Replace obsolete react context

### LINKED ISSUE #584: [addons-react] Following React 16.3+ new Context API ?

pablopalacios:
Closes #584 

### DIFF
--- a/packages/fluxible-addons-react/src/FluxibleComponent.js
+++ b/packages/fluxible-addons-react/src/FluxibleComponent.js
@@ -2,32 +2,21 @@
  * Copyright 2015, Yahoo Inc.
  * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
-import { Component, cloneElement } from 'react';
-import { func, object, node } from 'prop-types';
+import { Component, cloneElement, createElement } from 'react';
+import { object, node } from 'prop-types';
+import { FluxibleProvider } from './FluxibleContext';
 
 class FluxibleComponent extends Component {
-    getChildContext() {
-        return {
-            getStore: this.props.context.getStore,
-            executeAction: this.props.context.executeAction
-        };
-    }
-
     render() {
-        return cloneElement(this.props.children, {
-            context: this.props.context
-        });
+        const { children, context } = this.props;
+        const childrenWithContext = cloneElement(children, { context });
+        return createElement(FluxibleProvider, { context }, childrenWithContext);
     }
 }
 
 FluxibleComponent.propTypes = {
     children: node.isRequired,
     context: object.isRequired
-};
-
-FluxibleComponent.childContextTypes = {
-    executeAction: func.isRequired,
-    getStore: func.isRequired
 };
 
 export default FluxibleComponent;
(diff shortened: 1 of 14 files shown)


pablopalacios:
@redonkulus I gave it a try today. I just realized that there is another breaking change: since we are not using the old context anymore, we can't have in our API the feature of injecting other context values besides fluxible ones:

**connectToStores**

```javascript
// before
function connectToStores(Component, stores, getStateFromStores, customContextTypes)

// after
function connectToStores(Component, stores, getStateFromStores)
```

**provideContext**

```javascript
// before
function provideContext(Component, customContextTypes)

// after
function provideContext(Component)
```

redonkulus:
I think that makes sense as long as users can still add to the context. Otherwise, we would have to recommend an alternative method. @lingyan can you remember often we add custom context values, I know plugins do, seems like this would be a common thing. 

I would recommend testing this in the doc site to see if any other issues come up. 

Unrelated, but it would be good to migrate to import / export in these files too. This would help with tree shaking as well.

pablopalacios:
@redonkulus Another breaking change while updating fluxible-router: NavLink use to have a possibility to use a custom logger through contextTypes as well. I removed that so I could consume fluxible-context easily inside the class. Would that be fine?

pablopalacios:
> I think that makes sense as long as users can still add to the context. Otherwise, we would have to recommend an alternative method. @lingyan can you remember often we add custom context values, I know plugins do, seems like this would be a common thing.
> 
> I would recommend testing this in the doc site to see if any other issues come up.
> 
> Unrelated, but it would be good to migrate to import / export in these files too. This would help with tree shaking as well.

I didn't get what makes sense. I see 3 possibilities here:

1) Remove `customContextTypes` support (users will have to come up with their own react context to provide fluxible plugin values);
2) Convert `customContextTypes` object to `pluginsKeys` to an array. (Still breaking change, but we keep the feature);
3) Keep as it is (no breaking changes, but it is unnecessary to get an object with prop-types);

For the commonjs issue, I've just create another PR (#664 )

pablopalacios:
@redonkulus I've added back the `customContextTypes` feature in `provideContext`. Now, one can pass an array of plugins keys to inject its content into the state provided by `FluxibleProvider`. It is still a breaking change but the migration should be simple (just extract the keys from customContextTypes into an array).

redonkulus:
@pablopalacios Sounds good. Can you start a MIGRATATION.md doc in this branch to keep track of the breaking changes and what the code changes will be?

pablopalacios:
@redonkulus I've removed `FluxibleConsumer` and `useFluxible` from this PR since it is out of scope. So here you will find only the necessary steps to replace the obsolete react context. I created a initial document for the migration guide, but I did this in another PR (#665) so it doesn't block the code development. 

By the way, would it be possible to rebase the 1.x-dev branch on top of master?

redonkulus:
> By the way, would it be possible to rebase the 1.x-dev branch on top of master?

@pablopalacios Ok I've rebased and pushed, please update your fork.

pablopalacios:
@redonkulus, @lingyan any comments here?

redonkulus:
Sorry, I will get this merged and released today.

redonkulus (approved):
Changes look good to me, thanks for all the work you are doing!

Issue #439, opened 2016-05-06
RouteStore::dehydrate serializes too much information


jlecomte:
Here is what I get as a result of the Fluxible Router's RouteStore dehydrate function:

```
{
    "context": {
        "dispatcher": {
            "stores": {
                "RouteStore": {
                    "currentNavigate": {
                        "transactionId": 123456789,
                        "url": "\u002Ffoo",
                        "route": {
                            "method": "GET",
                            "path": "\u002Ffoo",
                            "handler": function StoreConnector(props, context) {
                                React.Component.apply(this, arguments);
                                this.state = this.getStateFromStores();
                                this._onStoreChange = null;
                                this._isMounted = false;
                            },
                            "action": function xyz(context, route, done) {
                                [tons of application code]
                            },
                            "name": "foo",
                            "url": "\u002Ffoo",
                            "params": {
                                ...
                            },
                            "query": {
                                ...
                            }
                        },
                        "error": null,
                        "isComplete": true
                    },
                    "routes": null
                },
                ... more stores ...
            }
        }
    },
    "plugins": {}
}
```

I don't think that `handler` and especially `action` should be serialized for each route, since the latter one can be very large.

jlecomte:
This issue blocks us because we are trying to not transpile much on the server now that we can use Node 6, and therefore, ES2015 code gets through, and Safari barfs on it when the page loads.

mridgway:
How else would `handler` and `action` be loaded and attached to the route for client side? I'm surprised this is only breaking in node 6. 

Typically only dynamic routes are dehydrated and using `action` and `handler` in with dynamic routes will not work properly because the dehydrated functions will not have the correct closure context that they need. This is where supplying a `getAction` and `getComponent` plugin to be able to retrieve handlers and actions by strings would be more beneficial.

jlecomte:
I am almost 100% certain that it did not work that way 6 months ago when I migrated Norrin to Fluxible.

Our client side bundle contains the definition for the route handler. It should not need to be inlined.

juandopazo:
@mridgway I think we figured this out and you can close this issue 😄

redonkulus:
@juandopazo what was the issue?

juandopazo:
We're serializing the route handlers when we could be referencing them as strings. I'm figuring out how exactly to get the handler from the string.

redonkulus:
Ok let us know if you have any questions.

PR #670, opened 2020-11-15
fluxible-router: remove route from dehydrated state

### LINKED ISSUE #522: dehydrated value of RouteStore contains `action` value [fluxible-router v1]
When dehydrate RouteStore on server side, the value contains `action` of RouteStore.
I’m in troubled that transpired code is exported to client side by server side setting.

I want the dehydrated value of RouteStore to be abstracted as before.

### RouteStore
```javascript
RouteStore.withStaticRoutes({
  index: {
    path: '/',
    handler: IndexComponent,
    action: async (context) => {
      await Promise.all([
        context.executeAction(FooAction, {}),
        context.executeAction(BarAction, {}),
      ]);
    },
  }
})
```

### v0.4.x dehydrated value
```javascript
{
  "RouteStore":{
    "currentUrl":"\u002F",
    "currentNavigate":{
      "transactionId":2953444313308840,
      "url":"\u002F",
      "config":{},
      },
    "currentNavigateError":null,
    "isNavigateComplete":true,
    "routes":null
  }
}
```

### v1.0.0 dehydrated value
```javascript
{
  "RouteStore":{
    "currentNavigate":{
      "transactionId":6733112344216282,
      "url":"\u002F",
      "config":{}
      "route":{
        "path":"\u002F",
        "handler":function StoreConnector(props, context) {
          React.Component.apply(this, arguments);
          this.state = this.getStateFromStores();
          this._onStoreChange = null;
          this._isMounted = false;
        },
        "action":async (context) => {
          await Promise.all([
            context.executeAction(FooAction, {}),
            context.executeAction(BarAction, {}),
          ]);
        },
        "method":"GET",
        "name":"index",
        "url":"\u002F",
        "params":{},
        "query":{}
      },
      "error":null,
      "isComplete":true
    },
    "routes":null
  }
}
```

### Steps to reproduce the behavior

  1. Transpile javascript with different babel settings on server and client.
    - server: `['env', { targets: { node: 'current' } }]`
    - client: `['env', { targets: { browsers: [...] } }]`
  2. Set async function in `action` to create RouteStore.
  3. Embed the serialized dehydrate value in the template and render it.
  4. Open in browser that not compatible with `async/await`.
    - `SyntaxError: Unexpected token '=>'`
    - ``Error: `rehydrate` called with a non-object. Ensure that the parameter passed to rehydrate is a state object produced by a dehydrate call.``

pablopalacios:
Closes #522 (and unsolved #439)

@redonkulus we're facing the issues mentioned in the links above in my company. The summary is:

Given that:
- RouteStore is dehydrating the route handler and action,
- handler and action are nodejs compatible code and
- [RouteStore doesn't need this data to re-hydrate](https://github.com/yahoo/fluxible/blob/fluxible-router-1.x/packages/fluxible-router/lib/RouteStore.js#L167).

The following problems occur:
- We are transferring way more data than we need for re-hydration (harming performance)
- The dehydrated state may even break the application (either by incompatible syntax or by a not found reference)

Previously we were doing some crazy workarounds on our application and I thought it was specific to us. But after checking #522, #439 and fluxible-router codebase, I realized that we could have this fixed in the package itself.

### DIFF
--- a/packages/fluxible-router/lib/RouteStore.js
+++ b/packages/fluxible-router/lib/RouteStore.js
@@ -151,8 +151,9 @@ var RouteStore = createStore({
     dehydrate: function () {
         // no need to dehydrate this._prevNavigate, because it will always
         // be null on server request
+        var currentNavigate = Object.assign({}, this._currentNavigate, {route: null});
         return {
-            currentNavigate: this._currentNavigate,
+            currentNavigate: currentNavigate,
             routes: this._routes
         };
     },
--- a/packages/fluxible-router/tests/unit/lib/RouteStore-test.js
+++ b/packages/fluxible-router/tests/unit/lib/RouteStore-test.js
@@ -35,6 +35,7 @@ describe('RouteStore', function () {
                 expect(state.currentNavigate.transactionId).to.equal('first');
                 expect(state.currentNavigate.url).to.equal('/foo');
                 expect(state.currentNavigate.method).to.equal('get');
+                expect(state.currentNavigate.route).to.equal(null);
                 expect(state.routes).to.equal(null);
             });
         });


lingyan (approved):
LGTM, especially since `RouteStore doesn't need this data to re-hydrate.`

pablopalacios:
Thanks for the review :). @lingyan, @redonkulus could we have a 1.x release of fluxible-router?

redonkulus:
@pablopalacios fluxible-router@1.7.0 published

redonkulus:
@pablopalacios we probably want to cherry-pick this into the 2.x branch as well

pablopalacios:
@redonkulus yup yup. But I don't know which branch I should base on. fluxible router on master is outdated if compared to the branch we are using for fluxible-addons-react. Since we already released a beta version of both packages, could we merge addons branch on master and keeping moving from there?

redonkulus:
ya its a little funky right now, I'd say just cherry pick the change into the fluxible-addons-react branch we are using for now. Soon we will be able to merge that into master when we are more comfortable with the changes and migration path.

PR #681, opened 2021-04-29
Examples: create up-to-date minimal example 

pablopalacios:
It works with React 17 :)


redonkulus:
@pablopalacios long time no see 😄 ... thought you were in production and done with the migration.

Should we remove or try to refactor the old examples to work with 17?

pablopalacios:
I tried to refactor the old examples a long time ago (then I got frustrated because I was not moving forward, then I had vacations, then end of the year then...). 

The build system there is quite old and it's a global one for all examples. So, if you touch one example you are touching all them. So I thought it would be better to keep the old examples in one folder and start migrating them one by one. 

But, to be honest, besides the `fluxible-router` and the `react-router` examples, I would try to add more modern ones instead of refactoring the other ones. For example a code splitting with loadable components and a razzle integration would be quite cool. I've done them already in my company and it would not be so hard to port it to here.

I would also do a simpler example focused on fetchr integration as well. The ones that we have IMO are quite complicated and not so helpful for the people maintaining legacy systems nowadays.

Regarding this example, I did it so I can show new devs in my company how fluxible works in the simplest way as possible. Since we have a quite big and complex application it's not easy for them to learn.

redonkulus:
Fair and excellent points. I'm completely ok with removing the old examples and adding simpler ones that highlight different areas of the fluxible architecture. I'll review this asap.

redonkulus (approved):

PR #682, opened 2021-05-01
addons-react: add helpful error messages to improve debugging

pablopalacios:
1. Since `provideContext` now accepts an array instead of an object, it would be better if we tell users that they forgot to update it. The previous behavior could lead to hard to find bugs due to missing plugin functionality.
2. Also, since connected components must have the fluxible context in its react context, it would also be better if we tell users that they forgot to wrap their components with `provideContext` or `FluxibleComponent`. The previous behavior would fail silently, making it hard to debug. 


redonkulus (approved):
Nice! Good changes here, looks good to me.

PR #687, opened 2021-05-04
examples: update fluxible-router example

pablopalacios:
Many things happened here:

- Updated to use latest `fluxible-router` and `fluxible-addons-react`
- Updated all the other dependencies
- Removed unused dependencies
- Removed dependency on Python for browser-only
- Removed dependency on babel plugins
- Simplified webpack configurations
- Use `Helmet` to manage page title
- Converted class components to function components

@redonkulus The only thing that I didn't update was purecss. This I will leave to you since I have the feeling that you know it better than me :p


redonkulus (approved):
Another great PR, thank you for doing all this work, amazing!

PR #696, opened 2021-05-15
Improve ref support on connectToStores (and remove refs from other components)

pablopalacios:
*addons-react/connectToStores*
- refactored tests
- removed `wrappedElementRef` ref
- add support for `React.forwardRef` with `{ forwardRef: true }`

*router/handleHistory*
- removed `wrappedElement` ref

*router/handleRoute*
- replaced refs test with a test that checks if props are passed correctly
- removed `wrappedElement` ref

*addons-react/batchedUpdatePlugin*
- refactored tests to not rely on refs

*addons-react/provideContext*
- removed `wrappedElementRef` ref

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "lint": "eslint .",
     "test": "lerna run lint --since --stream && lerna run test --since --stream"
   },
-  "dependencies": {},
   "devDependencies": {
     "@babel/cli": "^7.11.6",
     "@babel/core": "^7.11.6",
@@ -43,6 +42,7 @@
     "pre-commit": "^1.0.7",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
+    "react-test-renderer": "^16.0.0",
     "shelljs": "^0.8.0",
     "sinon": "^9.0.1",
     "yargs": "^16.0.0"
--- a/packages/fluxible-addons-react/UPGRADE.md
+++ b/packages/fluxible-addons-react/UPGRADE.md
@@ -51,13 +51,18 @@ If you were using `provideContext` to provide other context data not
 related to fluxible itself, you will need to provide your own
 solution to achieve the same result as before.
 
+#### Ref support removed
+
+This hoc does not include `wrappedComponent` ref anymore.
+
 ### connectToStores
 
-`connectToStores(Component, stores, getStateFromStores, customContextTypes)` -> `connectToStores(Component, stores, getStateFromStores)`
+`connectToStores(Component, stores, getStateFromStores, customContextTypes)` -> `connectToStores(Component, stores, getStateFromStores, options)`
 
 Since the new React API doesn't rely on PropTypes anymore, there is no
-need to specify customContextTypes to extract plugins context from the
-fluxible context since all plugins will be available.
+need to specify the `customContextTypes` param to extract plugins
+context from the fluxible context. All plugins component context are
+available.
 
 **Before:**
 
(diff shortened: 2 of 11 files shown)


redonkulus:
Are refs not needed anymore? I forget why they were there in the first place.

pablopalacios:
That was the question that I was making myself all the time since the only place that I saw it was in the tests. 

Buuut, I took a second look and I think we needed the refs so we could have access to the wrapped component instance or DOM in someway. Maybe one could have a connected input and need to focus it though a ref and maybe this ref mechanism that we implemented was allowing it. 

If that is the case, probably `React.forwardRef` would be the way to go. If it's not, I would assume that it was only created for tests or when people would interact more often with components instances through their methods or something like this.

I'll make some experiments here and see what I find. But maybe something that I said already rang some bells on you... Any other idea?

pablopalacios:
Yup, I think I've figured out. The current refs allow the parent component to have access to the wrapped component through the `connectToStores` HOC. One use case is the `focus` one that I mentioned above.

I did the experiments with all variations of inputs and refs. You can check it live [here](https://csb-4qkbv.netlify.app/) and the source [here](https://codesandbox.io/s/fluxible-refs-4qkbv?file=/src/App.js).

As you can see, we currently don't support `React.forwardRef` components. This is bad because we require users to transform their components in classes (which may not be possible/desired sometimes). Besides that, the current implementation is quite mouthy and not documented. If one wants to focus a wrapped component, this is what must be called from the parent component:

```javascript
// master branch (uses deprecated string ref)
this.ref.current.wrappedElement.focus()

// dev branch
this.ref.current.wrappedElementRef.current.focus()
```

My proposal is to follow `react-redux` approach. In their `connect` function, it's possible to pass an `options` object with `forwardRef: true` (see their docs [here](https://react-redux.js.org/api/connect#forwardref-boolean)). 

What they do is a conditional version of what is presented in the [React docs](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components) regarding refs and hocs. You can check the relevant code from them [here](https://github.com/reduxjs/react-redux/blob/master/src/components/connectAdvanced.js#L479), but basically what they do is, if the user sets `forwardRef`, they wrap the hoc with `React.forwardRef`, otherwise they return the hoc straight away.

Said that, my proposal is to adopt the same approach as them. The relevant part of `connectToStores` would look roughly like this:

```javascript
function connectToStores(Component, stores, getStateFromStores, options) {
  class StoreConnector extends React.Component {
    render() {
      const { fluxibleRef, ...props } = this.props;
      return <Component ref={fluxibleRef} {...props} {...this.state} />;
    }
  }

  if (options.forwardRef) {
    return React.forwardRef((props, ref) => (<StoreConnector fluxibleRef={ref} {...props} /> ));
  }

  return StoreConnector;
}
```

The user code could be:

```javascript
// CustomInput.js
export default connectToStores(CustomInput, stores, getStateFromStores, { forwardRef: true })

// App.js
const App = () => {
  const ref = React.useRef(null);

  return (
    <div>
      <CustomInput ref={ref} />
      <button onClick={() => ref.current.focus()}></button>
    </div>
  );
};
```
As you can see, we wouldn't have any ref at all in `connectToStores`. We would only trigger the possibility to forward refs if the user explicitly asks for it (and the user would be responsible for complying with refs rules instead of us trying to figure out if we should create/forward a ref or not). Finally, the call to `focus` in the parent component would be just as if `CustomInput` was a regular `input` component: `ref.current.focus()`.

By the one, I would only keep/fix the ref support on `connectToStores` since it's a hoc that is used in many lower components. Regarding `provideContext`, `handleHistory` and `handleRoute`, since they are all top level components that are created only once, I see no reason to keep ref support there.

redonkulus:
Thank you for the detailed analysis. Your recommendation makes sense to me and follows what other projects have done. I'm fine with this unless @mridgway or @lingyan have any reservations. I do not remember all the use cases for using ref's via connectToStore but making it an option and pass through prop makes sense to me.

As for `provideContext`, `handleHistory` and `handleRoute` I'm not sure what their use case was for `refs`.

pablopalacios:
Oh, I learned a lot during this analysis, so, no need to say thanks :)

I've updated the branch with my proposal so you can take a better look on how things could be.

pablopalacios:
Sure! The only problem is that we must not insert the ref when it's a functional component but still insert refs in forwardref components. I think `react-is` can help here. I'll give it a try soon.

pablopalacios:
@redonkulus I tried it and meh... It didn't work.

The problem is that `react-is` does not work with components, only with elements. @mridgway even tried to raise this issue on react itself (https://github.com/facebook/react/issues/12882) and @ljharb from enzyme tried to get a patch approved but it's been ignored since 2019 (https://github.com/facebook/react/pull/15349).

This means that there is no clean way to check if the component that must be wrapped by `connectToStores` is "ref-able" or not. If we don't do this check and pass a ref unconditionally, React will throw a warning for every function component (which would be super annoying):

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```
My opinion is that React documentation is a little bit incomplete regarding HOCs and forward refs. It assumes that all wrapped components will be "ref-able" components. 

If we don't want to play dirty, the only way that I see to circumvent this issue is by using some kind of flag that the user controls as proposed before. One thing that we could do differently is, instead of using `options.forwardRef` to determine if we should apply `React.forwardRef` or not, we could use it to check if we should forward the ref. Seems dumb and confusing but this is how it would roughly like:

```javascript
function connectToStores(Component, stores, getStateFromStores, options) {
    class StoreConnector extends ReactComponent {
         // ...
         render() {
             const ref = options.forwardRef ? this.props.fluxibleRef : null;
             return <Component ref={ref} />
         }
    }
    // ...
    return React.forwardRef((props, ref) => <StoreConnector fluxibleRef={ref} />);
}
```

In this way we would always have a forwardRef component (dismissing the problem of conditionally applying it as discouraged by react docs) but we would leave to the user to decide if a ref should be forward or not (eliminating the need to introspect the component). One benefit that I see here is that the API will likely survive react updates: if they decide to create any other ref-able component, we don't have to update any thing on our side.

redonkulus:
> we could use it to check if we should forward the ref.

I like this approach and can future proof us against React changes.

pablopalacios:
@redonkulus updated :)

redonkulus (approved):
This is great, I like the changes, seems clear to me. Thanks for doing the hard work!

redonkulus:
@pablopalacios new beta versions published to npm

PR #705, opened 2021-06-08
feat(addons-react): create withFluxible and useFluxible utilities

pablopalacios:
These were the missing pieces to provide fluxible context to components without relying on `contextType`.

Usage:

```javascript
import { withFluxible, useFluxible } from 'fluxible-addons-react';

// Functional components
const Component = () => {
    const { executeAction, getStore } = useFluxible();
    return null;
} 

// Class components
class Component extends React.Component {
    render() {
       const { executeAction, getStore } = this.props.context;
       return null;
    }
}

export default withFluxible(Component); 
```
I also renamed `FluxibleComponent` to `FluxibleComponentContext` so it doesn't clash with `fluxible.FluxibleContext` class.
 


redonkulus (commented):
Do you want to update examples and docs in this PR as well?

pablopalacios:
I will do examples in another one since it will probably trigger way more changes. But I'll send the docs.

pablopalacios:
@redonkulus I added the documentation entries for functions I created. While doing this, I also realized that the overall docs must be completed updated (I missed that in someway). I'll create another PR for the whole docs update.

packages/fluxible-addons-react/docs/api/withFluxible.md
```diff
@@ -0,0 +1,24 @@
...
+  render() {
+    const { context } = this.props;
+
+    return (
+      <button onClick={() => context.executeAction(myAction)} />
+    );
+  }
+}
+
+export default withContext(Component);
```
redonkulus, line 23: Should this be withFluxible?

redonkulus (commented):
Looks good, just one question

packages/fluxible-addons-react/docs/api/withFluxible.md
pablopalacios: Fixed.

redonkulus (approved):
Thanks again. Changes look good. Let’s make sure the upgrade guide and site docs are updated to reflect these changes.

PR #683, opened 2021-05-02
fluxible-addons-react: improve packaging

pablopalacios:
I did small packaging improvements to decrease final package size:
- Add `es` and `commonjs` builds to the final package (it should be tree-shakeable now);
- Remove `src` from final package;
- Use `babel-runtime` to decrease final modules size by not inlining babel helpers on each file;
- Convert `FluxibleComponent` to function to decrease final module size (4.2kb -> 0.7kb);

**Before**
```bash
npm notice === Tarball Contents === 
npm notice 1.1kB dist/batchedUpdatePlugin.js     
npm notice 766B  src/batchedUpdatePlugin.js      
npm notice 8.0kB dist/connectToStores.js         
npm notice 3.1kB src/connectToStores.js          
npm notice 2.4kB dist/createElementWithContext.js
npm notice 1.1kB src/createElementWithContext.js 
npm notice 4.2kB dist/FluxibleComponent.js       
npm notice 700B  src/FluxibleComponent.js        
npm notice 4.6kB dist/FluxibleContext.js         
npm notice 1.2kB src/FluxibleContext.js          
npm notice 1.5kB dist/index.js                   
npm notice 534B  src/index.js                    
npm notice 6.4kB dist/provideContext.js          
npm notice 2.0kB src/provideContext.js           
npm notice 937B  package.json                    
npm notice 2.5kB CHANGELOG.md                    
npm notice 338B  CONTRIBUTING.md                 
npm notice 1.7kB LICENSE.md                      
npm notice 1.6kB README.md                       
npm notice 2.7kB UPGRADE.md                      
npm notice === Tarball Details === 
npm notice name:          fluxible-addons-react                   
npm notice version:       1.0.0-beta.0                            
npm notice package size:  9.5 kB                                  
npm notice unpacked size: 47.3 kB                                 
npm notice shasum:        6a89b0a24e31f1f6fcf443edf28409b53b85610e
npm notice integrity:     sha512-z1Oludsi8OQI0[...]Nmb7hat97SjUw==
npm notice total files:   20      
```

**After**
```bash
npm notice === Tarball Contents === 
npm notice 914B  dist/es/batchedUpdatePlugin.js      
npm notice 1.1kB dist/lib/batchedUpdatePlugin.js     
npm notice 4.3kB dist/es/connectToStores.js          
npm notice 4.9kB dist/lib/connectToStores.js         
npm notice 1.2kB dist/es/createElementWithContext.js 
npm notice 1.5kB dist/lib/createElementWithContext.js
[...]

PR #710, opened 2021-06-13
router: improve packaging and remove inherits

pablopalacios:
`fluxible-router` packaging works in the same way as `fluxible-addons-react` now. Summary of changes:

- converted cjs modules into es modules;
- added babel runtime as dependency;
- `dist` now builds cjs and an es package;
- merged addons and lib folders since package is now tree-shakeable;
- removed inherits in favor of standard es classes;
- updated docs;


redonkulus:
Thanks for this. One thing to keep in mind is that we had already start some refactoring of this library in this branch https://github.com/yahoo/fluxible/tree/fluxible-router-1.x

The current 1.x npm alpha had this change in it. However, we have done releases from this branch over that one. I forget, but did you incorporate the changes from that branch in this one or did we lose it?

packages/fluxible-router/docs/api/HistoryWithHash.md
```diff
@@ -27,7 +27,7 @@ By default, the hash fragments are just url paths.  With `HistoryWithHash`, you
 
 ```js
 var handleHistory = require('fluxible-router').handleHistory;
```
redonkulus, line 29: should we change these to import?

packages/fluxible-router/package.json
```diff
@@ -24,10 +27,10 @@
     }
   ],
   "dependencies": {
+    "@babel/runtime": "^7.14.0",
```
redonkulus, line 30: Should this be a devDependency at the repo root?

pablopalacios:
> Thanks for this. One thing to keep in mind is that we had already start some refactoring of this library in this branch https://github.com/yahoo/fluxible/tree/fluxible-router-1.x
> 
> The current 1.x npm alpha had this change in it. However, we have done releases from this branch over that one. I forget, but did you incorporate the changes from that branch in this one or did we lose it?

Yeah, it's a little bit messy now. But I've checked again and all the changes from that branch were applied here ;)

packages/fluxible-router/package.json
pablopalacios: @babel/runtime is a runtime dependency so it can't be devDependency. And we need to have on each package.json since otherwise it would not be installed when one do `npm install fluxible-router`.

packages/fluxible-router/docs/api/HistoryWithHash.md
pablopalacios: I've updated the examples to use modern es ;)

packages/fluxible-router/package.json
redonkulus: whats the reason for needing it to be a runtime dependency?

redonkulus:
Ok that other branch will just be removed once we merge this branch into master. Thanks for keeping things in sync.

packages/fluxible-router/docs/api/README.md
```diff
@@ -1,29 +1,21 @@
...
+  handleHistory,
+  handleRoute,
+  navigateAction,
+  NavLink,
+  RouteStore,
+  HistoryWithHash
+} from 'fluxible-router';
 ```
-
-## Addons
```
redonkulus, line 17: why was this removed?

pablopalacios: Since `HistoryWithHash` is part of the main package now, I kept the same doc structure as for the other functions/classes. So an entry was added on line 8 for it.

pablopalacios: By the way, since the package is now tree-shakeable, there is no need to keep the addons in a separated folder. Also, when `fluxible-router` was changed to use babel, the import for this package got broken:

- v1: `import { HistoryWithHash } from 'fluxible-router/addons'`
- v2: `import { HistoryWithHash } from 'fluxible-router/dist/addons'`
- This PR:  `import { HistoryWithHash } from 'fluxible-router'`

packages/fluxible-router/package.json
pablopalacios: Oh, it's the same as we did for `fluxible-addons-react`. I explained it here: https://github.com/yahoo/fluxible/pull/683#issuecomment-830857300

packages/fluxible-router/package.json
```diff
@@ -2,13 +2,16 @@
   "name": "fluxible-router",
   "version": "2.0.0-beta.12.0",
   "description": "Routing for Fluxible applications",
-  "main": "index.js",
+  "main": "./dist/lib/index.js",
```
redonkulus, line 5: I think you can just remove the dist folder and just having everything off of "lib" since its all under "src" now.

packages/fluxible-router/package.json
redonkulus: oh right, totally forgot your great explanation. do we need to track the hardcoded babel version to remove that later? any follow ups with this approach we need to keep an eye on later?

packages/fluxible-router/package.json
pablopalacios: Oh, my fault, this lib is not the same as the previous lib. I'll rename it to cjs to make it clear.

pablopalacios: Done.

packages/fluxible-router/package.json
```diff
@@ -2,13 +2,16 @@
   "name": "fluxible-router",
   "version": "2.0.0-beta.12.0",
   "description": "Routing for Fluxible applications",
-  "main": "index.js",
+  "main": "./dist/cjs/index.js",
```
redonkulus, line 5: Do you need a module property for the es modules code now too?

pablopalacios: Yes! You are completely right!

packages/fluxible-router/package.json
pablopalacios: We could do as react-redux: https://github.com/reduxjs/react-redux/blob/master/.babelrc.js#L16. They read the version from package.json removing the possibility to have mismatched versions between the installed one and the consumed one. 

It's nice, but, in my opinion, giving that we don't update this package too often, we could keep it simple (as we did on addons-react).

Regarding future, it's nice to keep the package version updated since we can benefit more from newer utilities or see them completely vanish when browsers evolve. But the same is valid for all the other babel packages. So I would say that the only thing to keep in mind is, whenever we use update babel deps, we should update this one as well (and remember to update babelrc to point to the right version).

But if prefer, I can prepare another commit that reads the version from package.json as well. What do you say?

redonkulus: I sort of like the automatic version update that react-redux does. Its one less thing to remember when upgrading babel, because the version will get stale as new versions come out. I think you could do that in a follow up PR for both packages.

redonkulus (approved):

redonkulus:
Released in fluxible-router@2.0.0-beta.13.0

PR #730, opened 2021-08-26
chore(deps): use routr@3

redonkulus:
@pablopalacios Pushing this up so you can see the failed unit test, looks like the changes in path-to-regexp could lead to issues for apps:

```
  1) navigateAction
       should include query param on route match:

      AssertionError: query added to route payload for NAVIGATE_START{"method":"get","path":"/","name":"home","url":"/?foo=bar&a=b&a=c&bool","params":{},"query":{"foo":"bar","a":"c","bool":""}}: expected { foo: 'bar', a: 'c', bool: '' } to deeply equal { Object (foo, a, ...) }
      + expected - actual

       {
      -  "a": "c"
      -  "bool": ""
      +  "a": [
      +    "b"
      +    "c"
      +  ]
      +  "bool": [null]
         "foo": "bar"
       }
```

You can see the error here https://github.com/yahoo/fluxible/pull/730/checks?check_run_id=3437749374

### DIFF
--- a/packages/fluxible-router/package.json
+++ b/packages/fluxible-router/package.json
@@ -32,7 +32,7 @@
     "fluxible-addons-react": "^1.0.0",
     "hoist-non-react-statics": "^3.3.2",
     "prop-types": "^15.7.2",
-    "routr": "^3.0.0"
+    "routr": "^3.0.1"
   },
   "peerDependencies": {
     "fluxible": "^1.0.0",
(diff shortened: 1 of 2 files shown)


pablopalacios:
After updating it to routr 3.0.1 (which fixes a bug with arrays), the `null` from this line must be replaced with an empty string: https://github.com/yahoo/fluxible/blob/master/packages/fluxible-router/tests/unit/navigateAction-test.js#L92.

PR #741, opened 2021-10-30
docs(examples): add a stales props example

pablopalacios:
@redonkulus after going deep on how react-redux works and all the issues that they faced regarding stale props, I was curious to know how fluxible would handle the situation. Since we are not doing any thing special to guarantee that subscriptions are happening from top to bottom (so, child components subscribe before parent components since this is how React works), a simple todo application could crash if no proper measures are taken.

So I did this example to display the issue just to document how one can handle this case.

By the way, I also did a version with the previous implementation with the legacy react context just to check how it would behave. The results are the same.

More info regarding stale props:

- https://blog.isquaredsoftware.com/2018/11/react-redux-history-implementation/
- https://kaihao.dev/posts/Stale-props-and-zombie-children-in-Redux

### DIFF
--- a/examples/stale-props/README.md
+++ b/examples/stale-props/README.md
@@ -1,0 +1,32 @@
+# Stale props example
+
+This example shows how to handle the stale props issue. The stale
+props issue happens when the following conditions are met:
+
+1. A child component relies on its props to retrieve data from the store
+2. The child component subscribes to the store before its parent.
+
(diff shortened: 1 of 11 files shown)


redonkulus (approved):
Thanks for adding this, good idea to document this properly. I wonder if we should update the website with a link to this example or some guide for users.

PR #736, opened 2021-10-09
build: add node 16 to ci build matrix

pablopalacios:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #742, opened 2021-10-31
test: replace mocha with jest

pablopalacios:
@redonkulus I applied the jest codemod and fixed everything that was wrongly transformed.

I also migrated the tests to cjs since it works better with jest.

This PR will allows us to move forward with #736.


redonkulus (approved):
Changes look good, do you have more info on import not working with jest?

pablopalacios:
To make imports work with jest, there are some extra steps that must be done: https://jestjs.io/docs/ecmascript-modules. But as the jest docs states, mocking is not a clean thing with esm.

Another alternative would be to just use babel for everything. But given that most of the fluxible packages don't need any transpilation, it just feels wrong to do all the hackery or the babel integration to support esm.

By the way, even for the packages that we support esm, since at test runtime we would be importing the cjs modules, it just doesn't make any sense to work with esm there (IMO).

Perhaps, as soon as IE11 is not supported anymore, we will only need babel for jsx transpilation. But it could also be a better deal if we migrate everything to typescript.

PR #752, opened 2021-12-12
fix(addons-react): prevent fluxibleRef to leak in connected components

pablopalacios:
I'm migrating a huge app to the newer fluxible-addons-react/fluxible-router. It has many enzyme tests that checks if the result of `getStateFromStore` is correct. Example:

```js
describe('MyComponentConnected', () => {
  it('pass correct props to inner component', () => {
    const wrapper = mount(<MyComponentConnected />);
    const innerComponent = wrapper.find(MyComponent);

    const expectedProps = {
       foo: 'bar'
    };

    expect(innerComponent.props()).to.deep.equal(expectedProps);
  });
});
```
Due to the lack of attention, we were forwarding the internal `fluxibleRef` prop to the inner component. The test above would fail since MyComponent props would be:

```js
{
  fluxibleRef: null,
  foo: 'bar',
}
```

`fluxibleRef` is a implementation detail that we use internally to forward the user ref between the component created with `forwardRef` and `StoreConnector` ((see: https://github.com/yahoo/fluxible/blob/2ba3719c1dd5418b8af8c3ba2241e581e3e4c076/packages/fluxible-addons-react/src/connectToStores.js#L77) ). It should never be passed to the user component as `fluxibleRef` (only as ref). 

This patch fixes this issue by only passing the rightful props to the inner component.

### DIFF
--- a/packages/fluxible-addons-react/src/connectToStores.js
+++ b/packages/fluxible-addons-react/src/connectToStores.js
@@ -62,11 +62,12 @@ function connectToStores(Component, stores, getStateFromStores, options) {
         }
 
         render() {
-            const storeState = getStateFromStores(this.context, this.props);
+            const { fluxibleRef, ...props } = this.props;
+            const storeState = getStateFromStores(this.context, props);
 
             return createElement(Component, {
-                ref: this.props.fluxibleRef,
-                ...this.props,
+                ref: fluxibleRef,
+                ...props,
                 ...storeState,
             });
         }
@@ -75,10 +76,12 @@ function connectToStores(Component, stores, getStateFromStores, options) {
     StoreConnector.contextType = FluxibleComponentContext;
 
     const forwarded = forwardRef((props, ref) =>
-        createElement(StoreConnector, {
-            ...props,
-            fluxibleRef: options?.forwardRef ? ref : null,
-        })
+        createElement(
+            StoreConnector,
+            options?.forwardRef
+                ? Object.assign({ fluxibleRef: ref }, props)
+                : props
+        )
     );
     forwarded.displayName = `storeConnector(${
         Component.displayName || Component.name || 'Component'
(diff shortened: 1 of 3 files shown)


packages/fluxible-addons-react/package.json
```diff
@@ -15,7 +15,7 @@
     "precover": "npm run dist",
     "prepublish": "npm run dist",
     "pretest": "npm run dist",
-    "test": "../../node_modules/.bin/jest -c ../../jest.config.js",
+    "test": "../../node_modules/.bin/jest -c ../../jest.config.js packages/fluxible-addons-react/",
```
redonkulus, line 18: Was this change intentional?

redonkulus (commented):
Overall, changes look good to me, aside from that one comment

packages/fluxible-addons-react/package.json
pablopalacios: Yup. If we don't pass the package folder (using the repo root as reference), jest will run all tests for all packages when we do a npm test inside the package folder.

redonkulus: Forgot about this, but why didn't we have to do this with any other package?

pablopalacios: Actually we do, I just was a little bit lazy to add it for the other ones as well. I'll send a commit.

pablopalacios: Done ;)

redonkulus: How are you executing the tests? If you run `lerna run test --scope fluxible-addons-react` then lerna will automatically change the cwd to that folder, does this work for you?

pablopalacios: If I'm working only in one package, I `cd` to it's folder and run `npm t`. But if I'm changing many packages at a time, then I run `npm t` in the root folder. So, basically, I pretend that lerna does not exist.

Your command works but it's not as easier to remember as just `npm t`, right? But if you prefer, I can undo it, no problem ;)

redonkulus: Ya lets back that change out for now since its unrelated to this PR anyways. We can think about the best way to tackle that separately.

pablopalacios: Done! But I knew it was not a dejavu: https://github.com/yahoo/fluxible/pull/647#discussion_r486614345 :D

redonkulus (approved):

PR #793, opened 2023-03-08
fix: lint

snyamathi:
```
46:33  error  Unknown property '_target' found  react/no-unknown-property
```


lingyan (approved):
good catch

